PR #2, opened 2020-08-15
Use bincode for serialization

marc-casperlabs:
This one should be less controversial, as the topic has already been discussed. This is mostly about recording the findings.

[Rendered](https://github.com/marc-casperlabs/ceps/blob/use-bincode-for-serialization/text/0002-use-bincode-for-serialization.md)

### DIFF
--- a/text/0000-use-bincode-for-serialization.md
+++ b/text/0000-use-bincode-for-serialization.md
@@ -1,0 +1,131 @@
+# Use Bincode for serialization
+
+## Summary
+
+[summary]: #summary
+
+RFC PR: [casperlabs/rfcs#0000](https://github.com/casperlabs/rfcs/pull/0000)
+
+Currently there are multiple serialization formats in use inside the Rust node's storage and various networking protocols. This proposal outlines why they can (and should) all be replaced with [Bincode](https://crates.io/crates/bincode).
+
+## Motivation
+
+[motivation]: #motivation
+
+Serialization is a core necessity for the CasperLabs node, as it has to transfer large amounts of data safely across the network and store it. Any overhead results in reduced performance, while any security issues will instantly expose the node to attackers, as they are on a public network.
+
+Bincode is one of the most efficient Rust-native serialization formats available. Until recently it was unclear whether it held up in terms of security, but since these questions seem to have been [answered](https://github.com/servo/bincode/pull/346), it is likely the best available candidate today.
+
+As a result, serialization and deserialization should become a non-issue for any Rust-only part of the software, as it can leverage the [serde](https://crates.io/crates/serde) and [bincode](https://crates.io/crates/bincode) crates to great effect.
+
+## Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+Two core problems that need to be solved through serialization are storage and networking: How are we going to persist data on disk, and how are we going to send it across the network. Storage typically requires a very efficient binary representation and invariance across multiple versions of the software, but can rely on trusting its saved data. Networking, while also sensitive to serialized size, most importantly requires safety from malicious inputs.
+
+The [Bincode](https://crates.io/crates/bincode) format used to have no answers for these questions before, but recently they were [addressed in issue #345 of the project](https://github.com/servo/bincode/issues/345#issuecomment-673025338). Given a correct configuration, the format should be:
+
+* Invariant over endianness
+* Stable across minor versions
+* Secure against untrusted inputs (no undefined bevavior)
+* Protecting against memory-exhaustion (provided it is configured correctly)
+* Space efficient (with varint encoding enabled)
+
+**Varint encoding**: Variable length integer encoding encodes integers in different sizes depending on their size. For example, a `u32` might take up 1-5 bytes depending on its values, as opposed to the usual 4 bytes by encoding small values with a single byte.
+
+Some alternatives considered directly were JSON, Msgpack, CBOR; see the alternatives section.
+
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+The code changes are minimal due to the fact that the serialization method is already behind a layer of abstraction in the form of the `serde::Deserialize` and `serde::Serialize` traits.
+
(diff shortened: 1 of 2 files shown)


marc-casperlabs:
> Well, speed was the blocker for using it inside EE, and still is as far as I'm aware.

I do think the speed is not going to be an issue here - without profiling I am very wary to assume so, although very willing to be convinced otherwise by evidence =).

My estimate is that the limiting factor for the node as a whole in terms of CPU will usually be the event loop which cannot be parallellized, or huge computations like running the execution engine. Those computations have to be run on a different thread anyway to not block the event loop.

IO (and thus serialization and deserialization) already runs on a different thread for the same reason as well. I do not think we are going to saturate a full core with just IO, even with transport cryptography added, before we hit bandwith limits. Similarly I don't think we can saturate all cores, given that our most computing intensive task (EE) is hard to parallellize.

> Not mentioning it and also not mentioning the ToBytes/FromBytes as a potential alternative seems a bit misleading in my opinion.

I do not think it's an alternative at all since there is no Serde support, but that certainly is worth mentioning in the CEP - misleading was not my intention here, it just did not occur to me to be even a remotely viable option. Just to be sure: We would have to implement and test all serialization/deserialization for structs manually?

> As an aside, the rendered link in the initial comment is broken now (maybe due to the repo being renamed or made public)

Thanks, forgot that one when the repo was renamed from rfcs to ceps. Fixed now!

marc-casperlabs:
Finally mentions `FromBytes`/`ToBytes`, @Fraser999 . Any other blockers for this?

fizyk20:
Have you pushed the latest changes? I can't see the `ToBytes`/`FromBytes` mention, and I still see the typo that Fraser pointed out, even though his comment is marked as resolved. Is this maybe somehow related to the repo visibility change?

marc-casperlabs:
> Have you pushed the latest changes? I can't see the ToBytes/FromBytes mention, and I still see the typo that Fraser pointed out, even though his comment is marked as resolved. Is this maybe somehow related to the repo visibility change?

Yes, my repo is currently a commit ahead. I've contact GitHub support about restoring the ties to the repo that were severed when visibility changed.

marc-casperlabs:
@fizyk20 @Fraser999 I've updated the branch this PR is based on with the latest commits. Please have a look.

Fraser999 (approved):

fizyk20:
There is still one suggestion from Fraser that is marked as resolved, but doesn't seem to be applied (and it looks like a typo correction) - is that intended? (I mean this: https://github.com/CasperLabs/ceps/pull/2#discussion_r473184264)

Fraser999:
@EdHastingsCasperLabs 
> In the current implementation I believe that would lead to replacing the rmp_serde based logic used in storage with ToBytes / FromBytes impls.

Yes - that's correct.

I don't disagree with your preference, but I just want to clarify whether or not you intend to use `To/FromBytes` in the context of messaging as well as storage?  That would leave only config file handling still using serde I think.

I also think we should look to update `ToBytes` as the first step towards making this change before adding many more impls.  It really shouldn't always be allocating to a new `Vec`.  I think something like the following would be far more efficient:
```rust
fn to_bytes<W: std::io::Write>(&self, writer: W) -> Result<(), Error>;
```
and we could get rid of `into_bytes()` altogether then.

marc-casperlabs:
> On reflection, I want to pump the brakes on this CEP. I'm losing confidence that either message pack or bincode are sufficiently version proof / deterministic for our long term storage needs. I think that determinism within minor version changes just isn't strong enough.

I am not sure I understand why though. This in turns means only major version changes will change the protocol, which is the definition of a major version change, AKA breaking changes!

What I do not get is the scenario that everyone is afraid of so badly. Is it


1. There being a change in serialization despite promises that that would not happen? No worries, we can pin the version.
1. The format being changed in a major version? Cargo will protect us here.
1. The crate being removed / altered? Not possible due to `crates.io`.
1. Format change as above AND a major security vulnerability being discovered AND the maintainer refuses to release a fix for the major version branch we are currently using? We can fork the crate and fix it, its open source!

> It does chafe a bit to contemplate, but I'm starting to settle on the notion that given the intended lifespan of this network, to absolutely avoid nondeterministic serialization regardless of any external factor, we should instead rely upon our own specific ABI format (protocol version bound) not just for global state but also for general storage.

We can introduce a tag for future proofing that indicates the serialization format, but I wondering where this non-determinism should come from? The only thing that is currently non-deterministic, to my knowledge, is `HashMap`s key order, which I believe we do not even use right now.

> In the current implementation I believe that would lead to replacing the rmp_serde based logic used in storage with ToBytes / FromBytes impls.

Not using serde where possible would be a huge mistake, hopefully I do not have to elaborate on that (but I will if pressed ;))

@Fraser999

> I also think we should look to update ToBytes as the first step towards making this change before adding many more impls. It really shouldn't always be allocating to a new Vec. I think something like the following would be far more efficient:

```rust 
fn to_bytes<W: std::io::Write>(&self, writer: W) -> Result<(), Error>;
```

> and we could get rid of into_bytes() altogether then.

I fear we will end up reimplementing parts of serde, just worse, with a lot more effort!

goral09:
> I fear we will end up reimplementing parts of serde, just worse, with a lot more effort!

so far, according to our benchmarks, `ToBytes`/`FromBytes` is faster than serde ;)

marc-casperlabs:
> so far, according to **our benchmarks**, ToBytes/FromBytes is faster than serde ;)

Emphasis mine =)

Even if, we're concerned about long term format stability here, not pure performance, right?

EdHastingsCasperLabs:
> What I do not get is the scenario that everyone is afraid of so badly. 

I don't pretend to speak for "everyone" or existential fears others may be feeling, however my main concern stems from our practice of generating hashes used for identity from the serialized bytes of structs, including the blocks and deploys that constitute our chain, which should remain readable for the lifetime of the platform even in the face of major / breaking changes.

marc-casperlabs:
I guess in this case we consider this thoroughly discussed. It seems a way to formalize `FromBytes`/`ToBytes` is a way forward, which will be discussed in a separate CEP.

In the meantime, I'd like to thank everyone for their input on this. =)

PR #3, opened 2020-08-16
Hash-indexed object store

marc-casperlabs:
This changes some of the foundations of the node itself, but may very well be worth it in the medium run.

[Rendered](https://github.com/marc-casperlabs/rfcs/blob/hash-index-object-store/text/0003-hash-indexed-object-store.md)

PR #5, opened 2020-08-21
Node joining syncing

goral09:
[Rendered view](https://github.com/goral09/ceps/blob/node-joining-syncing/text/0005-syncing.md)


afck:
I can't see the rendered view, probably because I don't have permissions to read your fork.

goral09:
> I can't see the rendered view, probably because I don't have permissions to read your fork.

Fixed.

text/0005-syncing.md
```diff
@@ -0,0 +1,114 @@
...
+**Necessary work:**
+
+1. Start a node with the trusted hash
+2. Ask peers for the initial block from the linear chain.
+3. Synchronize Global State at the post-state hash of the initial block.
+4. Validate correctness of the Global State with the post-state hash from the initial block.
+
+
+---------------------
+Once whole linear chain is "consumed" and state is built-up, we need to synchronize DAG(s) of still active eras.
```
afck, line 97: By "whole linear chain", you mean the segment from the initial to the most recent block?
(And maybe back as far as the unbonding period requires?)

text/0005-syncing.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Synchronizing DAG(s) ###
+
+#### Variant A (quick-n-dirty) ####
+
+Once node has built up its Global State and has necessary eras started it could rely on already existing mechanism of syncing nodes – i.e. it could transition into an *active* node (not a validator) and wait for new messages. Upon receiving a message for which it doesn't have all the necessary dependencies it will already start synchronizing.
```
afck, line 103: It sounds like that could delay passing messages between honest validators by a lot, because one of the intermediate hops has to download the whole DAG to verify that one message. I guess in the long run, only variant B is viable?

text/0005-syncing.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### Variant B
+
+If we want to fully sync the DAG(s) before transitioning from the synchronizing to an active node modes we need to ask peers for their latest DAG tips and compare it to our own view. Requires:
+
+* adding a new variant to `HighwayMessage::GetTips` which should return node's `panorama`
+* joining node should send `GetTips` message to its peers for all active eras
```
afck, line 110: I'm wondering whether instead of trusting a random selection of peers, we could download all dependencies of the switch block.
For that, we'd ideally know the hash of the era's last _Highway block_, thought… not sure if we want to leak that into e.g. the first executed block of the next era or something…

afck (commented):
Looks good overall; just a few questions.

I think in the long run we'll want both variants B. But we'll need to decide what to do in the short term.

It would also be great if the document defined exactly what data a node in the network is expected to have available:
* The linear chain? How far back?
* Highway DAGs? How many eras?
* Global state? Only for the most recent block, or for several blocks?

text/0005-syncing.md
```diff
@@ -0,0 +1,114 @@
...
+
+1. It should gossip only valid messages.
+2. It should gossip messages from eras it's not bonded in.
+
+To achieve **1.** we need a knowledge about the whole DAG – otherwise we cannot validate messages. Receiver of an invalid message should punish its creator and also the sender. This means that if we are an active node in the current era and we are expected to gossip current-era messages, we need that era's whole DAG. Even if we are not a ***validator*** in the current era we need to receive and process every vertex of that era.
+
+Point **2.** needs a bit more context: our consensus is using eras to rotate validator set. Only validators bonded in the current era are allowed to produce messages in that era. Once a new era is started, there's an era tail – time when validators produce more messages in that era even though there's a new era running. Validator can equivocate in the past era, and while we have its stake in the PoS contract (until the unbonding period passes), we have to punish him for that (take its staked tokens). Slashing an equivocating validator needs a cryptographic proof – a pair of messages created by the validator that prove an equivocation.
```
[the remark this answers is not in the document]
goral09: Funny, I read both versions and didn't notice any changes.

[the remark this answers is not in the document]
goral09: Ah, now I see: _him_ -> _it_ etc.

text/0005-syncing.md
goral09: Yes, it's _quick-n-dirty_ variant that requires least effort to add the sync feature.

text/0005-syncing.md
goral09: The main reason to send `GetTips` to multilpe nodes is to parallelize the synchronization process. Node should still add only correct vertices to the DAG (since it already knows validators for active eras).

text/0005-syncing.md
goral09: I hope it won't be _that_ slow. The main contributor will be deploys that have to be downloaded…

text/0005-syncing.md
goral09: Depends on what variant we go with. If it's **Variant A** then it will be the whole chain from Genesis, which is the _initial_ like you're saying although that's different from Genesis only for nodes that already have _something_ in their Linear Chain storage.

text/0005-syncing.md
```diff
@@ -0,0 +1,114 @@
...
+
+1. Add `era_id` to `BlockHeader`
+2. Add `switch_block` flag to `BlockHeader`
+3. (optional) Add next-era validator set to the `BlockHeader` if`block.switch_block = true`.
+
+Normal way of operation of an active node is that Consensus component will send `FinalizedBlock` messages to the `BlockExecutor`. `FinalizedBlock` message has fields like:
+
+* `switch_block` flag,
+* `random_bit`
+* `era_id` (doesn't yet have but it should)
```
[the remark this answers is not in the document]
goral09: Great :)

text/0005-syncing.md
```diff
@@ -0,0 +1,114 @@
...
+**Necessary work:**
+
+1. Start a node with the trusted hash.
+2. Ask peers for the initial block from the linear chain.
+3. Synchronize all missing dependencies.
+4. Execute deploys from the linear chain in the order of them appearing in the linear chain.
+
+`BlockExecutor` sees whether it's a `switch_block` and whether it should run an auction for era `era_id + AUCTION_DELAY` . As a result of execution, it will produce a `Block` instance (that is a block of the linear chain) but also an event `NextEraValidators(finalized_block.era_id + 1, Set<Validator>)` that informs Consensus about the next era validators. Consensus component will use new validators (and random bits from blocks between booking block and key block of the next era) and create an instance of new era.
+
+Same mechanism can be used in a node that is catching up – it will use `Block`s from the linear chain to build up its Global State via `BlockExecutor` (we need a mechanism of turning `Block` into `FinalizedBlock` as that's the input that `BlockExecutor` understands) and create new eras.
```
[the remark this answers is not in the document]
goral09: I was thinking that maybe instead of using parent's `ProtoBlock` hash we could use block height. This should be the same on all nodes and can be used to get block's `pre_state_hash`.

Fraser999 (approved):
It'd be good to understand what components will be in play during the synchronisation phase.  I presume consensus wouldn't be required at least?

The reason I ask is that it's not clear to me whether we'd be better to aim for a separate `synchronizer` reactor which transitions to a `validator` one on completion of sync (similar to the `initializer` one now), or whether we just have a state flag in the validator reactor to indicate synching is in progress and which effectively disables the consensus and any other unneeded components during sync.

It's also a bit of a factor in the Variant B of DAG sync.  The `GetTips` message would be handled by the consensus component of a recipient.  But I'm wondering if the sender needs to have consensus running in order to make the requests and handle the responses?  Or could that be handled by a separate component in the `synchronizer` reactor and then used to initialize the consensus component on transition?

Perhaps these questions are outside the scope of this CEP though.

goral09:
> It'd be good to understand what components will be in play during the synchronisation phase. I presume consensus wouldn't be required at least?
> 
> The reason I ask is that it's not clear to me whether we'd be better to aim for a separate `synchronizer` reactor which transitions to a `validator` one on completion of sync (similar to the `initializer` one now), or whether we just have a state flag in the validator reactor to indicate synching is in progress and which effectively disables the consensus and any other unneeded components during sync.
> 
> It's also a bit of a factor in the Variant B of DAG sync. The `GetTips` message would be handled by the consensus component of a recipient. But I'm wondering if the sender needs to have consensus running in order to make the requests and handle the responses? Or could that be handled by a separate component in the `synchronizer` reactor and then used to initialize the consensus component on transition?
> 
> Perhaps these questions are outside the scope of this CEP though.

These are all good questions and I would say that they are implementations details. 

If I had to answer now though, I would say that having a separate `synchronizer` reactor may be the cleanest but `flag` might be the fastest :)

goral09:
> It'd be good to understand what components will be in play during the synchronization phase. I presume consensus wouldn't be required at least?

that will be:
* consensus component
* block executor + storage
* network layer/gossiper

 Now, whether we can reuse already existing ones or we will implement "leaner" versions depends on which variant we go with.

text/0005-syncing.md
goral09: As for **Variant B**, if we first synchronize only the Global State then we don't need the linear chain at all (of course we won't have any part of it that comes _before_ the initial hash). We will sync the Global State and then proceed to synchronize the DAG(s) – this process will create the linear chain as a side-effect of syncing (at least that was my idea for it).

text/0005-syncing.md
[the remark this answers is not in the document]
MParlikar: We didn't observe that all nodes had the same block height for a given block hash in the Scala node. What is different in Rust impl. that makes this so?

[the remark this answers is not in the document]
goral09: It's the linear chain – a chain of finalized blocks – and they all have the same block height. They had in Scala too (since all nodes have to finalize the same thing in the same order).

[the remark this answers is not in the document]
afck: That's true; unless there is an equivocation catastrophe — which we'd detect locally before finalizing any two blocks at the same height — the height should suffice.

afck (commented):
No blockers left from my side, except that I'd like to have a clear definition in this document (at least for B) of what data exactly (linear chain, global state, Highway state — and for how far into the past) a node is expected to have before it can join the network and start gossiping.

And — maybe less urgent — a clear definition of how we can verify all the downloaded data, starting from the single trusted hash: Which blocks contain the validator sets and PRNG seeds for which eras? Just in the global state or also in the header? Do we have to go back from era 4 to era 3 to learn the validators of era 6, or can we learn them from an era-4 block? Do we need to download global state for that or just the header?

goral09:
> It would also be great if the document defined exactly what data a node in the network is expected to have available:
> 
> * The linear chain? How far back?

1. Depends on whether we can synchronize Global State. If we do, then we have to only keep the Global State for the switch block of the oldest active era. 

> * Highway DAGs? How many eras?

2. I think that once era is unbonded we don't need to keep its DAG anymore. 

Although, currently we don't have any persistence at all.

> * Global state? Only for the most recent block, or for several blocks?

3. I think that point 1. answers that.

goral09:
I just realized that I am making an assumption here that unbonding period and auction delays are the same. Is that a problem? cc @osolmaz @AlexanderLimonov

afck:
> Depends on whether we can synchronize Global State.

I think we should either decide that here (at least for the short term), or add the definition (of what state you're expected to keep) for both options to the document.

text/0005-syncing.md
marc-casperlabs: For our projected gamma (charlie?) net, I would assume this is not an issue due to the small size? With low traffic, the Global State is small, and with shorter eras there's an upper bound on DAG size, I would assume.

fizyk20: Messages are gossiped, right? So even if someone starts downloading the whole DAG, others would still be gossiping the message and it could reach everyone else before this someone finishes... no?

goral09: I think that Andreas' meant that if one of the "infection targets" is a syncing node then the ability to gossip new vertices to active validators is limited since part of the traffic is going through nodes that won't gossip (syncing node cannot gossip since we must gossip only valid messages – i.e. ones that we can validate).

marc-casperlabs:
I like this in principle, especially Variant A, which in both cases seems the best approach *for now* to me: Low implementation effort while being reasonably easy to verify. I especially do not want to add a method of synchronizing global state right now, as there is some overlap here with [CEP0003](https://github.com/CasperLabs/ceps/pull/3), which deals entirely with syncing data, which we should either reject or ratify before making any long-term decisions on this.

To me it seems of the three proposed criteria, "secure", "correct" and "fast", we should only concern ourselves with correct. "secure" will have to be entirely handled by social consensus and "fast" is just not a concern at the moment while there is a chance we will be able to get "fast" for free.

To me, that leaves us with "correct", arguably the hardest ;).

As the CEP is a bit light on implementation proposal at the moment, could something like this work (marking new implementations in bold)?:

1. We create a **new reactor** similar to the existing `Initializer`. Instead of starting a chain from genesis, it is passed a block hash to start with. I will call it the **`Joiner`**, because the name "synchronizer" has been used too often elsewhere.
2. The joiner reactor **downloads the linear chain** and verifies it. Note that we're not download the DAG, as variant A will do this for us during operation anyway.
3. The joiner reactor **executes all deploys starting from genesis** to verify the chain.
4. We find out if we have reached the current era. If not, abort (?) [is this necessary]. When retrying, we will have a lot of things already in our local storage and a better chance to catch up.

We now have an up-to-date global state and blockchain, with no era DAGs. We can now launch the regular validator reactor.

text/0005-syncing.md
fizyk20: Okay, it will be slightly limited for a moment, but we never infect just one node, right? And the rest will keep infecting others, so in general the setback shouldn't be too big.

At least from one node. If we have multiple syncing nodes, it might get worse.

osolmaz:
> I just realized that I am making an assumption here that unbonding period and auction delays are the same. Is that a problem?

We should not assume they are the same.

afck (approved):
Approving; I think we agree on everything.
But I'd still prefer if [this](https://github.com/CasperLabs/ceps/pull/5#issuecomment-679978460) were made precise and added to the document.

text/0005-syncing.md
AlexanderLimonov: Wouldn't Variant B also cause sync of everything in everyone's privately observed DAGs, as written, possibly going back to genesis?

goral09: Only the active, not-yet-unbonded eras' DAGs will be sync'd. That's because Consensus will create only eras that are created by the blocks in the linear chain and those are finalized blocks that should be the same for all the nodes in the network.

ashok-cl:
@goral09 @EdHastingsCasperLabs Looks like we can close this as it is approved or are you waiting to resolve [this ](https://github.com/CasperLabs/ceps/pull/5#issuecomment-679978460) and clarify implementation details as suggested by @marc-casperlabs [here](https://github.com/CasperLabs/ceps/pull/5#issuecomment-680033333)?

MParlikar (approved):

PR #6, opened 2020-08-24
Handling equivocation catastrophes

fizyk20:
[Rendered view](https://github.com/fizyk20/ceps/blob/equivocations/text/0006-equivocation-catastrophes.md)


text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,111 @@
...
+
+Here we will just summarize the exact steps to be performed when an equivocation catastrophe happens.
+
+1. An equivocation catastrophe is detected. Once a single validator detects it, we can expect all the validators to detect it eventually as well (even if just due to them receiving the evidence from the validator that detected it first).
+2. Consensus gets suspended, though validators continue to gossip. The users (validator nodes' owners) receive notice about the catastrophe and start the process of social consensus.
+3. The users agree about the last correct block, the transactions that need to be replayed on top of it and the set of equivocating validators (exchanging evidence if necessary).
+4. A chainspec for the restart is manually constructed and agreed upon. This includes:
+    - the initial set of validators for the restarted network which excludes the equivocators,
+    - the initial global state hash for the restarted network
+5. The network is being restarted using the chainspec constructed above.
```
goral09, line 74: I think that after detecting an equivocation and (socially) coming to consensus on which block (from the Linear chain) is the last trusted one you can just use the same mechanism as for the node sync/join described in my document (CEP-0005).

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,111 @@
...
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+Here we will just summarize the exact steps to be performed when an equivocation catastrophe happens.
+
+1. An equivocation catastrophe is detected. Once a single validator detects it, we can expect all the validators to detect it eventually as well (even if just due to them receiving the evidence from the validator that detected it first).
+2. Consensus gets suspended, though validators continue to gossip. The users (validator nodes' owners) receive notice about the catastrophe and start the process of social consensus.
+3. The users agree about the last correct block, the transactions that need to be replayed on top of it and the set of equivocating validators (exchanging evidence if necessary).
+4. A chainspec for the restart is manually constructed and agreed upon. This includes:
```
MParlikar, line 71: What is the risk here around 'tweaking' or 'messing with' this chainspec?  I don't know if there are any good solutions, seems like this needs to be tightly accounted for via governance rules?

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,111 @@
...
+    - the initial set of validators for the restarted network which excludes the equivocators,
+    - the initial global state hash for the restarted network
+5. The network is being restarted using the chainspec constructed above.
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+The single drawback is that this proposal requires manual intervention in the case of an equivocation catastrophe. An ideal situation would be if the network could handle these automatically, but this is most probably impossible.
```
MParlikar, line 80: Governance should tightly define the process to address an equivocation catastrophe, this reduces the amount that the process can be muddled with.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,111 @@
...
+The information necessary for restarting the network can (and probably should) be included in the chainspec. The "restart-chainspec" could be distributed among the validators the same way the initial chainspec for the network was.
+
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+Here we will just summarize the exact steps to be performed when an equivocation catastrophe happens.
+
+1. An equivocation catastrophe is detected. Once a single validator detects it, we can expect all the validators to detect it eventually as well (even if just due to them receiving the evidence from the validator that detected it first).
+2. Consensus gets suspended, though validators continue to gossip. The users (validator nodes' owners) receive notice about the catastrophe and start the process of social consensus.
```
afck, line 69: Actually, we might want to consider more than just the validators to be the "users" here. (Vlad Zamfir made that point a few months ago.)

Social consensus can't be defined precisely, of course, but e.g. in a scenario where _all_ validators go bad, others could still decide to continue running the system, with a completely new set of validators. Or if 90% of validators go bad, others who already have bids in the auction contract could be added to the new validators. Or even if 50% equivocated, the community — the users, exchanges, etc. — might not trust the other 50% anymore and want to continue running with a different set of validators.

So maybe point 3 should allow, in addition to removing equivocators, to:
* remove validators without slashing them, and
* add new validators, and
* possibly even adding new stakes to the auction contract?

text/0006-equivocation-catastrophes.md
afck: I think once the equivocation catastrophe has happened, in a sense all bets are off.
Because if we defined a strict governance process, with strict rules and a strict quorum, then too many actors of that off-chain process could be malicious, too. Anything that can be precisely defined can be implemented, and is subject to equivocation catastrophes itself.

Social consensus allows the system to survive even such a situation, just like it can survive an equivocation catastrophe: Because _anyone_ can continue the chain, with _any_ set of validators, and what matters is only whether the public accepts that as the "real" new chain.

text/0006-equivocation-catastrophes.md
afck: Except that we'll (in the long run) need to provide tools to make the required changes to the global state (e.g. removing validators) and use the new global state as the starting point, instead of only the post-state of the last good block.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,111 @@
...
+
+The single drawback is that this proposal requires manual intervention in the case of an equivocation catastrophe. An ideal situation would be if the network could handle these automatically, but this is most probably impossible.
+
+## Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+As mentioned above, automatic handling of equivocation catastrophes is probably impossible (as it would effectively mean constructing a consensus algorithm that exceeds proven theoretical limitations), so manual intervention is really the single option of dealing with such situations.
+
+An alternative to social consensus would be to have a designated party deal with the situation. However, such a solution would go against the goal of creating a truly distributed system. In this alternative, the party responsible for handling equivocation catastrophes would effectively single-handedly control the network, which is something we want to avoid.
```
afck, line 88: I totally agree!
But our _recommended_ process could still be to wait for CasperLabs to make a restart proposal (propose the new starting block and validator set), carefully check that, and if it's agreeable, go with it.
Because that's effectively also our default process for protocol updates: CasperLabs proposes them, and hopes that the validators and the community will accept them.

We can argue that that's not centralization, because if the community decided at any point to stop following us, and agreed on a different process for updating the software and handling equivocation catastrophes, they could do that.

The downside for us would be that we'd have to do a good job with those recommendations: We'd need to have someone "on-call" 24 hours/day, who could quickly resolve such situations.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,111 @@
...
+
+Not handling equivocation catastrophes at all would mean the end of the network once such a situation happens. Thus, even though we expect the probability of an equivocation catastrophe happening to be low (and we are making efforts towards making it as low as possible), the costs in the event of it happening are great enough to justify guarding against it.
+
+## Prior art
+
+[prior-art]: #prior-art
+
+There have been cases of blockchain projects introducing manual interventions in order to revert catastrophical events, for example the Ethereum DAO hack.
```
afck, line 96: It would be interesting to collect a few links to similar documentation of other projects. (But that would probably be a whole separate research story.)

You could link to [Vitalik's blog post](https://ethereum.github.io/blog/2014/11/25/proof-stake-learned-love-weak-subjectivity/), though.

Also, [this Confluence page](https://casperlabs.atlassian.net/wiki/spaces/EN/pages/146047313/Weak+subjectivity) is probably obsolete now. Not sure whether we archive or delete those…?

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+- Which option for choosing the last correct block do we pick?
+- Do we even require the last correct block, or do we just need to agree on an initial global state for the restarted network? Having the starting point for the new network in the form of a block would help in keeping the continuity of the blockchain and can be useful for history archiving purposes, but doesn't seem to be strictly required for keeping the functionality of the network.
+- A related question which will be addressed in a separate CEP is about the process of joining the network by new validators. A new validator will have to be able to join the correct, agreed upon instance of the network, and might have to refer to social consensus in order to get the correct initial data (the correct chainspec).
```
afck, line 104: That's CEP0005, is it? Would be good to link to it (or the PR) here.

text/0006-equivocation-catastrophes.md
goral09: Hmm, I thought that the point of social consensus was to agree on a point in the history at which there's no equivocation – i.e. is the last _safe_ block. Maybe that equivocation was caused by a bug and not an intentional attack etc.

Modifying the consensus' state, like getting rid of equivocators, is a different concern.

text/0006-equivocation-catastrophes.md
fizyk20: Good point. Completely redefining the set of validators should probably be possible. This actually would make the restart after an equivocation catastrophe even more similar to the initial launch of the network, where we also define an arbitrary set of validators.

Maybe it would even be best to define handling of equivocation catastrophes in terms of a generalization of the initial network startup process?

text/0006-equivocation-catastrophes.md
fizyk20: Looks like it :+1:

text/0006-equivocation-catastrophes.md
fizyk20: A good question. I would expect that someone messing with their chainspec locally would mostly hurt themselves, because their node software would expect a different initial state of the network than everyone else's. It could be a problem if a majority of the validators collude in order to make some changes to the chainspec which others didn't agree to... But that would probably mean them effectively creating a hard fork (or are there possibilities of messing with the chainspec that would go unnoticed by other nodes for some time and create problems with some delay? cc @afck  @goral09 )

text/0006-equivocation-catastrophes.md
afck: I'd say it's part of it. After an equivocation catastrophe, we _can't_ continue from the last good block without modification: The unmodified validator set already contains too many equivocators to continue.

text/0006-equivocation-catastrophes.md
afck: I agree: Changing the chain spec means creating a hard fork.
If the others are not following your fork, it's useless and not dangerous.
If the others _are_ following your fork, you successfully reached social consensus, and it's the accepted fork now.

text/0006-equivocation-catastrophes.md
goral09: By the "last good block" I meant one from before (any, some?) equivocations. If the equivocation was a result of software bug, patching software and starting over from the last good block is enough. If starting over would lead to equivocations again, then you're right – we cannot continue with equivocators as validators.

text/0006-equivocation-catastrophes.md
goral09: > Maybe it would even be best to define handling of equivocation catastrophes in terms of a generalization of the initial network startup process?

I like this idea. We could simply restart network from the "last trusted" era and use the new chainspec.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,111 @@
...
+
+The process of restarting the network could be very similar to the process of starting it. In its essence, it is just like starting the network from scratch, but with a few differences:
+
+- There is a different set of genesis validators, which is based on the bonded validators of the last correct block, with the equivocators removed.
+- There is a different initial global state: it is not empty anymore, but it is the post-state of the last correct block.
+- Other details like the protocol version can differ as well, but they are not of the main concern for the restart process itself.
+
+The information necessary for restarting the network can (and probably should) be included in the chainspec. The "restart-chainspec" could be distributed among the validators the same way the initial chainspec for the network was.
+
+## Reference-level explanation
```
osolmaz, line 62: Catastrophes are generally events that require validators to react very quickly, in a matter of hours. I think the reference-level implementation needs to include the time windows for different steps once the catastrophe gets acknowledged. E.g. maximum duration validators will wait to tally votes for step 3.

This way, there would be no liability of participating validators, if certain validators are too late to react, and try to object to the decision because they don't like to outcome.

I don't remember how fast the hard fork for the DAO hack was implemented, but if [this post](https://medium.com/@slacknation/dao-hack-timeline-823e5a18e894) is to be believed, Vitalik announced the recovery plan a bit under 8 hours following the attack.

text/0006-equivocation-catastrophes.md
osolmaz: I wouldn't say this is a drawback. That statement is a fact, and this CEP is proposed to remedy this issue in the first place.

Social consensus requires off-chain governance and manual intervention by default.

I'd say we have the following drawback: However we try to account for all possible scenarios with a legal process, there can always be one exception that this CEP has not accounted for. In that case, we would wait for it to play out, and eventually amend the CEP to account for the exception.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,111 @@
...
+The manual actions afterwards will consist of two main steps:
+
+1. Reaching social consensus about the block to be used as the new starting point,
+2. Restarting the network with the new starting point.
+
+We will go into more detail about these steps below.
+
+### Social consensus
+
+For the initial version of this CEP, we assume that we have some kind of platform that enables open communication between validators and CasperLabs. The exact details of what that platform will be (e-mail? Discord? a forum?) can be left to be determined in the future.
```
osolmaz, line 38: We should not even mention Discord. Possible options are email with signatures, or a forum/Matrix instance not controlled by a single validator.

The chosen platform should allow for storing, linking, referencing and displaying the whole public process and communication. We also should make sure that the content is in a format that can be displayed elsewhere. We should consider that people would refer back to these processes and disputes for the next 100 years.

text/0006-equivocation-catastrophes.md
afck: > similar documentation of other projects

Ah, that's probably covered by [RES-324](https://casperlabs.atlassian.net/browse/RES-324).

text/0006-equivocation-catastrophes.md
AlexanderLimonov: Mateusz makes an excellent point - there needs to be a differentiation in these procedures for handling software bugs and actual attacks, to the extent that we can distinguish them. We can't just burn millions of someone's dollars because we pushed a bad patch.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+The procedure for handling the equivocation catastrophes will have to consist of two main parts: what the software is expected to do in such a case, and what are the manual actions required to proceed.
+
+The obvious first thing that should happen in the case of an equivocation catastrophe is that the software should alert the users about the situation. The node software is aware of the FTTs of the blocks, so when it detects equivocations, it can notice if the total weight of the equivocators exceeds the FTT of a block. When the FTT of any block is exceeded, the users should be made aware of it.
```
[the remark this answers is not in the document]
AlexanderLimonov: @afck, does this imply option 1 needs serious clarification regarding the meaning of FTT? E.g., does it mean a _still_ finalized block, or the last finalized block before the catastrophe was observed (and by whom?)?

text/0006-equivocation-catastrophes.md
AlexanderLimonov: I don't think non-validator users should be explicitly involved (big dApps might be validators, anyway), but we need to set fairly precise expectations about what happens when a catastrophe hit, and let users write their own CEPs if they want to challenge it.

text/0006-equivocation-catastrophes.md
AlexanderLimonov: @afck, perhaps some consensus protocol issues alluded to with respect to formalizing a process could be circumvented via reliance on an _irrelevant_ third party, e.g., getting people into a Zoom room, where you can't equivocate on a vote without magic hacking skills and/or telepathy?

text/0006-equivocation-catastrophes.md
AlexanderLimonov: @MParlikar, would Mallika have time to write something up for us before this CEP resolves? Re: RES-234

text/0006-equivocation-catastrophes.md
AlexanderLimonov: I think this CEP could use a few more details about comms procedures, possible uses of Clarity UI, etc., otherwise we would just need to do all of that in another CEP. Alternatively, we could make this about selection of restart point/GS _only_ and do the restart itself in a later CEP? @fizyk20

AlexanderLimonov: Especially relevant since there appears to be a dependency on CEP-5, which isn't done yet.

text/0006-equivocation-catastrophes.md
afck: The means of communication is yet another question, yes. I'd recommend several independent communication channels, though; e.g. at least some forum _and_ Zoom, just in case: So that if Zoom is censoring you or even just having technical problems that prevent you from joining, you can tell the others via the forum.

But that's kind of orthogonal to the _process_: What do we _do_ once we're in the Zoom room? I think it's fine to not make any recommendation here. I think it's also fine to kind of promise that we'll make a proposal in such cases.

text/0006-equivocation-catastrophes.md
afck: Sure, it's not feasible to have a conference with tens of thousands of users. But if the meeting is among validators only and behind closed doors, that could hurt trust in the platform even more. So the expectation should be — and maybe that's something we should explicitly recommend — that the whole process and discussion is made public.

But from the _technical_ side, I think we also need to provide the means to overcome the absolute worst-case scenario: That all validators (and we ourselves) turn evil, and the community wants to replace them. I.e. some way to continue the network from a particular state, but with completely different validators and maybe some other state changes (slash everyone!) applied.

text/0006-equivocation-catastrophes.md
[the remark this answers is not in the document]
afck: It definitely has to mean a _still_ finalized block, because as you say, the "last observed" one is subjective.

But, if 40% equivocated, should we take the last block with 41% FTT? Or would we want some safety margin and require, say, 45%? Or, if we have a 33% FTT configured, would we now want 33% _of the remaining 60%_, which would mean an absolute FTT of 40% + 20% = 60%?
That's something we could also make some recommendation for.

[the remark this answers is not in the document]
AlexanderLimonov: Presumably we are ejecting the equivocating validators in a genuine (non-bug-related) equivocation attack, so it should be something that would be finalized if we excise them? "Highest ranked with highest mass among remaining validators" perhaps?

text/0006-equivocation-catastrophes.md
AlexanderLimonov: Well, the process can be made public without allowing public to actively participate in the _main_ communication channel.

The worst-case scenario could be provided for, but mostly it's a technical issue of restarting from a given state. We should have the feature, but we don't need to provision a governance process for a total schism between validators and the userbase. Something must have gone wrong far upstream of that scenario.

text/0006-equivocation-catastrophes.md
AlexanderLimonov: Well, it's not orthogonal, because being in a Zoom meeting (or any other channel where everyone sees everyone simultaneously) means that the points above about theoretical limits to consensus mechanisms no longer apply, since whole classes of misbehaviors are no longer possible (everyone heard you vote simultaneously, there's no equivocating). Of course, then you run into the limits imposed by social choice theory, but it's easier to think about voting rules relative to consensus algorithms.

afck: Good point, yes.
So are you suggesting to not recommend a particular process at all (other than having a call with all participants), or to define a precise process that basically implements a "consensus protocol" with perfect synchrony (i.e. live video call)?

text/0006-equivocation-catastrophes.md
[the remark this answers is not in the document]
afck: Yes, that should be equivalent with the remaining 60% above. So:
* If we go with the last block with 41% FTT, we basically assume that none of the remaining 60% has equivocated. (If we're all in a Zoom call and agree exactly on the set of equivocators, that's probably reasonable.)
* If we go with the last block with 60% FTT, that means requiring a 33% FTT _among the remaining 60%_, i.e. it would mean 33% if the equivocators had never existed.

text/0006-equivocation-catastrophes.md
fizyk20: Essentially, I want to keep this one as vague as possible, since the details could depend on the particular scenario and when it happens (which could influence what tools we have available, what is the main channel of communication with the community and various other things). The idea is to outline a general process, with the details being filled in when more is known about the specific case.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,111 @@
...
+
+A desirable goal for the initial global state would be to revert as few transactions from all the forks as possible, in order to limit the damage. Every transaction that was included in a block which appeared to be finalized for a time could have caused real world consequences (like, for example, a merchant sending ordered goods). Reverting any such transaction might potentially cause a loss to a user.
+
+In order to minimize losses, the following approach is proposed: first, choose the last finalized block as the starting point. If the equivocators caused multiple forks to appear, choose the tip of one of the forks. The exact criteria of choice will be left to the community to decide.
+
+After the block is chosen, take its global state and all transactions from the blocks in other forks and attempt to replay them on top of that global state. Since the order could matter, the community would have to agree on one. Call the resulting global state the "pre-initial state".
```
afck, line 46: This is something we need to provide the tools for, so that — without having to develop a whole new piece of software — the community can choose, define and configure that pre-initial state.
E.g. it could be defined by:
* a known block from before the equivocation,
* a list of blocks whose deploys should be replayed on top of that, and
* a list of system transactions to be run on top of that — e.g. to slash and/or refund (in case of bugs).

(But I guess the third point could be done in a separate step after the restart, as an upgrade point.)

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,111 @@
...
+
+After the users collectively decide upon the new initial state and the new initial set of validators, the network can be restarted.
+
+### Restarting the network
+
+The process of restarting the network would be very similar to the process of starting it. In its essence, it is just like starting the network from scratch, but with a few differences:
+
+- There is a different set of genesis validators - the new validators chosen by the community.
+- There is a different initial global state: it is not empty anymore, but it is the initial state prepared before.
+- Other details like the protocol version can differ as well, but they are not of the main concern for the restart process itself.
```
afck, line 58: …and we'd not start from height 0; in other words, we'd keep the linear chain from the original network.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,111 @@
...
+[reference-level-explanation]: #reference-level-explanation
+
+Here we will just summarize the exact steps to be performed when an equivocation catastrophe happens.
+
+1. An equivocation catastrophe is detected. Once a single validator detects it, we can expect all the validators to detect it eventually as well (even if just due to them receiving the evidence from the validator that detected it first).
+2. Consensus gets suspended, though validators continue to gossip. The users (validator nodes' owners) receive notice about the catastrophe and start the process of social consensus.
+3. The users agree about the new initial global state and the new initial set of validators. The new initial global state is manually constructed based on social consensus.
+4. A chainspec for the restart is manually constructed and agreed upon. This includes:
+    - the initial set of validators for the restarted network,
+    - the initial global state hash for the restarted network
```
afck, line 73: As mentioned above, I think this initial global state hash (including the replayed deploys and changed validator set) should not be given explicitly, but implicitly (just like the actual initial state hash when starting a _new_ network is also only given implicitly). Otherwise we need to provide special software to compute it.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,111 @@
...
+
+The first essential step will be determining the starting point for the network after the restart. The suggested approach here would be to agree upon the new initial global state, new set of validators and restart the network as if it was starting from scratch.
+
+A desirable goal for the initial global state would be to revert as few transactions from all the forks as possible, in order to limit the damage. Every transaction that was included in a block which appeared to be finalized for a time could have caused real world consequences (like, for example, a merchant sending ordered goods). Reverting any such transaction might potentially cause a loss to a user.
+
+In order to minimize losses, the following approach is proposed: first, choose the last finalized block as the starting point. If the equivocators caused multiple forks to appear, choose the tip of one of the forks. The exact criteria of choice will be left to the community to decide.
```
afck, line 44: > the last finalized block

That is not objective anymore, after the equivocation catastrophe, if only the output of our standard finality detector is taken into account.
What _would_ be objective, after syncing the full DAG, would be the last block still appearing finalized with a 1% FTT, but we'd need to add some code to find that block.

afck (approved):
Looks good to me! :+1: 

My remaining comments are about making sure we don't end up having to write a lot of additional code _after_ an equivocation catastrophe happens, i.e. that we implement all the tools needed for a restart ahead of time.

text/0006-equivocation-catastrophes.md
fizyk20: I think we will need special software to modify the initial global state, anyway (to replay chosen transactions and to modify the validator set). This software could just as well output the state hash after everything's done.

afck: I don't think it needs to be separate software, though. It could just be part of the node: Just like it currently computes the "initial" global state using the contracts configured in the chainspec, it should compute the new global state after such a restart, based on a _block_ hash (the last finalized block), a list of additional block and/or deploy hashes, and possibly some upgrade-like code.

text/0006-equivocation-catastrophes.md
fizyk20: Hmm... Would that have to be 1% FTT, or _any_ common FTT would be sufficient?
If it's not any FTT, then I guess the reason would be that FTTs exceeded by the weight of equivocating validators would not do. In such a case, wouldn't we have to choose the last block appearing finalized with some chosen FTT higher than the weight of equivocators?

afck: The way we currently calculate a block's FTT, equivocators are already excluded from any summit.
So once the catastrophe has happened and we've fully synced up our protocol states — let's call that the "final state" —, we'll all see the _f_ equivocators as faulty.
* Any block seen finalized with an FTT _> f_ has the property that nobody ever saw a conflicting block finalized (even before syncing up). The last such block might still be older than some blocks that _some_ observers saw finalized with FTT _> f_ (or above the configured threshold) before the catastrophe.
* There is a unique last block seen as finalized with FTT _> 0_ in that final state. That block is a descendant of all blocks that were ever observed finalized with FTT _> f_ by anyone.

If that's correct (it is, isn't it? :thinking:), I guess both would be canonical choices: The last block still seen with FTT _> f_, and the last block seen with FTT _> 0_ in the final state.
I think maybe the latter is preferable, because it means at least you know that whatever you observed with FTT _> f_ is still considered final. It gives you the guarantee that, if you locally require FTT 80% to consider something final, and fewer than 80% equivocate, even if the network considers that a catastrophe, your "finalized" blocks are still finalized after the restart.

fizyk20: > The last such block might still be older than some blocks that some observers saw finalized with FTT > f (or above the configured threshold) before the catastrophe.

I don't get this part. If FTT > f and the protocol states are fully synced, then all validators see the same blocks as finalized, and no finalization could be reverted, right? So why would the last finalized block in such a situation be older than some blocks some observers saw as finalized before the catastrophe?
Or is it a typo and you meant FTT < f in that last part?

afck: What I mean is that equivocations can now _reduce_ the FTT for already finalized blocks. If a block is finalized with 35%, and then 5% equivocate, it can happen that afterwards, that block's FTT is only 30%.

So it could happen that you already observed a block with 80% FTT, then 70% equivocated, so now that block looks like it has only 10%, even though some participants observed it as finalized with 80%.

I think it would be good if such a block would then be guaranteed to remain finalized.

fizyk20: Wait, so how does that work now? We still have some concrete FTT defined in the chainspec. Say it's 40%. We see a block finalized with this FTT, then 30% equivocate. Does that mean that the block is now no longer finalized with FTT 40%? Even though it's FTT hasn't been exceeded?

afck: Kind of. But we don't detect that and we continue considering it finalized anyway.
The idea is that it will almost certainly reach the configured FTT again, unless the equivocators actually exceed that limit. So instead of us doing something about that block not having sufficient FTT anymore, we just wait, and assume that eventually, everyone will see it finalized again.

fizyk20: Hmm... That sounds kind of unsafe. Could it be possible for malicious actors to propose blocks and vote in such a way that we'd have two competing blocks, one is finalized "temporarily", then they equivocate just enough for it not to be finalized anymore with the original FTT and eventually the other block is finalized with that FTT instead?

afck: No, we can't (I think :sweat_smile:).
The only way to finalize (with FTT _> 0_) a block conflicting with a block that anyone ever saw as finalized with FTT _f_ is if _f_ validators equivocate.

text/0006-equivocation-catastrophes.md
MParlikar: Here is the research: https://drive.google.com/file/d/1XuZ_3IcYbhg9hkPw2h6HH5rdkBsf2JS-/view?usp=sharing

MParlikar (changes requested):

AlexanderLimonov (commented):
Since I was re-added as a reviewer - my concerns remain the same as they were when originally expressed in the comments. I think that this CEP requires more work on defining the class of situations it is meant to deal with (just a reversion of a single block? percentage of equivocating mass? is there a solid definition of an "equivocation catastrophe?"), expected scenarios (*who* is attacking? could it be a bug? how do we know?) and remedies (what is the proposed main mode of communication?). 

Considering that we are doing this without any input from the validators who will have to do this, we either have to propose something specific and readily implementable that they can use without deliberation, or else resign ourselves to the validators inevitably having to come up with a specific solution themselves. 

Additionally, I think we are still quite far away from finalizing the infrastructure that would be critical to an emergency restart, such as hot joining and a method for efficiently sharing critical data (e.g., the chainspec and node config settings).

In conclusion, maybe this CEP should be re-written and re-considered at a later time, when we have more tools in place to work with and can receive feedback from validators who have built up some experience with operations.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,114 @@
...
+
+The first essential step will be determining the starting point for the network after the restart. The suggested approach here would be to agree upon the new initial global state, new set of validators and restart the network as if it was starting from scratch.
+
+A desirable goal for the initial global state would be to revert as few transactions from all the forks as possible, in order to limit the damage. Every transaction that was included in a block which appeared to be finalized for a time could have caused real world consequences (like, for example, a merchant sending ordered goods). Reverting any such transaction might potentially cause a loss to a user.
+
+In order to minimize losses, the following approach is proposed: first, choose the last safe block as the starting point. A safe block would be one that still appears finalized with some small FTT, for example 1%, after a full synchronization of the protocol state.
```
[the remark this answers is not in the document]
fizyk20: FTT is currently not a property of a block, but of the finality detector. The validators would have to recalculate consensus using a separate finality detector with a different FTT (which could be done automatically by the node software, I think).

text/0006-equivocation-catastrophes.md
fizyk20: Yes, I would imagine it would be a new chainspec.toml + any dependencies, so accounts.csv too. I believe that we should provide the necessary tools, yes (it is already mentioned in the latest version of the CEP).

fizyk20:
> I think that this CEP requires more work on defining the class of situations it is meant to deal with (just a reversion of a single block? percentage of equivocating mass? is there a solid definition of an "equivocation catastrophe?")

"Equivocation catastrophe" is defined in the Motivation section.
To be honest, I don't think it matters that much what the consequences are there. If we detect that too many validators are equivocators, we should act. Whether this means that a single block was reversed, or multiple ones, or something else happened - the general course of action would be the same and this is what this CEP attempts to outline.

> expected scenarios (_who_ is attacking? could it be a bug? how do we know?)

I'm not sure if these questions even _can_ be answered in advance. That's something we'll only know once the situation actually happens, and it is then when it should be discussed among the validators, at least in my opinion.

> and remedies (what is the proposed main mode of communication?).

This can also depend on many factors, and that's why I went with what Onur suggested and just listed properties this mode of communication should satisfy.

> In conclusion, maybe this CEP should be re-written and re-considered at a later time, when we have more tools in place to work with and can receive feedback from validators who have built up some experience with operations.

I would be fine with that, though I also think that it could serve as a very general outline of a course of action as it is now.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,114 @@
...
+
+We will go into more detail about these steps below.
+
+### Social consensus
+
+For the initial version of this CEP, we assume that we have some kind of platform that enables open communication between validators. The chosen platform should allow for storing, linking, referencing and displaying the whole public process and communication. We also should make sure that the content is in a format that can be displayed elsewhere.
+
+The first essential step will be determining the starting point for the network after the restart. The suggested approach here would be to agree upon the new initial global state, new set of validators and restart the network as if it was starting from scratch.
+
+A desirable goal for the initial global state would be to revert as few transactions from all the forks as possible, in order to limit the damage. Every transaction that was included in a block which appeared to be finalized for a time could have caused real world consequences (like, for example, a merchant sending ordered goods). Reverting any such transaction might potentially cause a loss to a user.
```
goral09, line 42: Should it be `at a time` instead of `for a time`?

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,114 @@
...
+
+In order to minimize losses, the following approach is proposed: first, choose the last safe block as the starting point. A safe block would be one that still appears finalized with some small FTT, for example 1%, after a full synchronization of the protocol state.
```
goral09, line 44: When I describe the order, I like to use a representation that makes it easier to follow the process – like an ordered list or bullet points.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+The procedure for handling the equivocation catastrophes will have to consist of two main parts: what the software is expected to do in such a case, and what are the manual actions required to proceed.
+
+The obvious first thing that should happen in the case of an equivocation catastrophe is that the software should alert the users about the situation. The node software is aware of the configured FTT, so when it detects equivocations, it can notice if the total weight of the equivocators exceeds the current FTT. When it is exceeded, the users should be made aware of it.
+
+In order to make sure that as many validators as possible become aware of all the equivocations eventually, the operation of the nodes shouldn't stop. Finalizing blocks should be halted, as finalization ceases to mean anything when it can be reverted, but the nodes should continue gossiping votes.
```
goral09, line 25: How will the validators know when all of them have reached the same protocol state? One where all equivocators are known all everyone has seen all of the consensus messages?

text/0006-equivocation-catastrophes.md
goral09, line 44: > after a full synchronization of the protocol state.

Will that happen in node or will validators have to synchronize in the external channels?

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,114 @@
...
+
+In order to minimize losses, the following approach is proposed: first, choose the last safe block as the starting point. A safe block would be one that still appears finalized with some small FTT, for example 1%, after a full synchronization of the protocol state.
+
+After the block is chosen, take its global state and all transactions from the blocks that were considered finalized at some point, but are no longer a part of the main branch of the chain, and attempt to replay them on top of that global state. Since the order could matter, the community would have to agree on one (the default one could be based on deploy timestamps, but the community could decide on some other ordering if deemed necessary). The deploys' TTL could also be expanded to account for the time spent reaching social consensus and allow them to still be valid afterwards. Call global state resulting from replaying these deploys the "pre-initial state".
```
goral09, line 46: > The deploys' TTL could also be expanded to account for the time spent reaching social consensus and allow them to still be valid afterwards. Call global state resulting from replaying these deploys the "pre-initial state".

Since deploy's TTL is also signed over, this requires changes to the node software that filters out expired deploys.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,114 @@
...
+
+As the last step, agree on the initial set of validators for the restarted network. The equivocators should be slashed, but there may be reasons to also exclude other validators, or even to start with a set of validators that has nothing in common with the validators from the stopped network. Once agreement has been reached, apply necessary changes to the pre-initial state: modify the validator stakes and change the entries in the auction contract. After this is applied, the resulting state will be the state used for the restart.
```
goral09, line 48: I think it's worth noting that if the equivocation catastrophe happened b/c of the software bug, initial set of validators may be unchanged from the pre-catastrophe one.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,114 @@
...
+
+After the users collectively decide upon the new initial state and the new initial set of validators, the network can be restarted.
+
+Note that tools would have to be provided that would facilitate the manual construction of a specific state. This could be realized in the form of node software constructing the state based on the provided hash of a block, list of block or deploy hashes defining deploys to be executed on top of that block, and possibly some additional upgrade code.
+
+### Restarting the network
```
goral09, line 54: This section should also include, mentioned earlier, set of deploys for replay/reinclusion.

text/0006-equivocation-catastrophes.md
```diff
@@ -0,0 +1,114 @@
...
+The agreed upon set of validators and the initial state would be included in the chainspec. The "restart-chainspec" (consisting of a new `chainspec.toml` along with any dependencies like `accounts.csv`) could be distributed among the validators the same way the initial chainspec for the network was. The files containing the new initial state could be distributed alongside the chainspec, or the nodes could download it from each other after they are started (which, of course, requires at least one node to possess the full copy of the state).
+
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+Here we will just summarize the exact steps to be performed when an equivocation catastrophe happens.
+
+1. An equivocation catastrophe is detected. Once a single validator detects it, we can expect all the validators to detect it eventually as well (even if just due to them receiving the evidence from the validator that detected it first).
+2. Consensus gets suspended, though validators continue to gossip. The users (validator nodes' owners) receive notice about the catastrophe and start the process of social consensus.
```
goral09, line 72: It's worth noting that validators may detect equivocation catastrophe at different moments so their final protocol states may differ – some of them may even finalize few more units after others have detected the catastrophe already.

text/0006-equivocation-catastrophes.md
fizyk20: I'll defer to native speakers here ;) But I believe `for a time` fits what I was trying to say (about a block that was finalized, but ceased to be after some time).

text/0006-equivocation-catastrophes.md
fizyk20: Good question. I wonder if a good solution here would be to stop creating new votes altogether, but keep gossiping the ones we have until the network looks saturated from our perspective?

text/0006-equivocation-catastrophes.md
fizyk20: See my reply to your comment above. I think it could happen in node, but maybe I'm missing something important.

text/0006-equivocation-catastrophes.md
fizyk20: The deploy hashes are mentioned - or do you mean that we should explicitly state that the deploy _bodies_ are needed as well?

text/0006-equivocation-catastrophes.md
fizyk20: > some of them may even finalize few more units after others have detected the catastrophe already

These would be the blocks whose finalization would be reverted by the equivocations, wouldn't they?

text/0006-equivocation-catastrophes.md
goral09: I meant in the list below – we somehow have to get that information into the node.

xcthulhu:
I have chatted with @fizyk20. In order to support recovering the network when too-many validators equivocate, we identified the following engineering action items:

  1. Support loading a global state from a file along with a genesis chainspec.  The chainspec will need an optional global state root hash. There will need to be a check for correctness of that state root hash.
  2. Support signing/checking a plain-text message with a validator secret. This is so that validators can safely identify their counter-parties as they communicate out-of-band.

Note the CEP does not specify the forum the validators will use to communicate.

CC @henrytill and @EdHastingsCasperLabs, since this entails work in the storage layer.

xcthulhu (approved):
The engineering requirements coming out of this CEP are clear. From my perspective we have actionable next steps we can proceed with.

MParlikar (approved):

PR #7, opened 2020-08-24
Daily gas price adjustment algorithm

osolmaz:
Design and specification for a daily gas price adjustment algorithm.

[Rendered](https://github.com/osolmaz/ceps/blob/daily-gas-price-adjustment/text/0007-gas-price-adjustment.md)


text/0007-gas-price-adjustment.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+This is to increase the speed of recovery from a catastrophic failure, by enabling as many users as possible to transact. We choose the value of `INITIAL_BASE_PRICE` low enough, so that resetting practically results in a floating price regime. That is, until the platform recovers from the failure and the algorithm catches up with the market price.
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+- The argument for this proposal favors *price stability* over *market efficiency*. In other words, we achieve stability by setting a price floor, and making users occasionally *pay higher* than the actual market price. If this argument is wrong, then the algorithm would be rendered useless.
+- The adjustment rate may fall too short to adapt to short term trends, like sudden increased usage due to a hyped dapp.
```
MParlikar, line 104: What happens to other applications in this instance?

text/0007-gas-price-adjustment.md
```diff
@@ -0,0 +1,135 @@
...
+CEP PR: [casperlabs/ceps#0004](https://github.com/casperlabs/ceps/pull/0004)
+
+Gas price is regulated algorithmically. A price floor exists and adjusts daily, based on fullnesses of blocks from the previous day, in order to absorb volatility and reduce congestion.
+
+## Motivation
+
+[motivation]: #motivation
+
+Gas prices are known to be subject to high volatility both in the short term and the long term. This is due to a combination of factors such as inelastic supply, [daily demand cycle](https://solmaz.io/2019/10/21/gas-price-fee-volatility/) and shifting trends.
+
```
MParlikar, line 16: Please include 'CasperLabs sees these issues with volatility as a key barrier to the adoption of public blockchain infrastructure by businesses.  It is also recognized that price stability alone is not enough, scalability is also a core concern in this arena.'

text/0007-gas-price-adjustment.md
```diff
@@ -0,0 +1,135 @@
...
+During times of high demand for a blockchain, one can observe prices change up to 10-100x in a matter of hours. A floating gas price and the absence of a reference point exacerbates the situation. Such volatility makes it difficult for dapp providers and users to plan ahead their expenses.
```
ashok-cl, line 17: `Such volatility makes it difficult for dapp providers and users to budget for their expenses`

text/0007-gas-price-adjustment.md
```diff
@@ -0,0 +1,135 @@
...
+- `ADJUSTMENT_RATE`: The rate at which `BASE_PRICE` can move up or down per day. Value: `10_000_000_000` (1%).
+- `DENOMINATOR`: Fractions will be represented as trillionths of 1. Value: `1_000_000_000_000`.
+- `BLOCK_GAS_LIMIT`: Maximum gas allowed in a block. Value: TBD.
+- `TARGET_FULLNESS`: The value which will determine whether `BASE_PRICE` should move up or down. Value: `650_000_000_000` (65%).
+- `MIN_FINALIZED_BLOCKS`: The minimum number of blocks that must be finalized in an adjustment period. Otherwise, a fail-safe is triggered. Value: TBD.
+
+### Transaction Fields
+
+- `PRICE_PREMIUM`: The price premium specified for a given transaction.
+- `MAX_GAS_PRICE`: The maximum gas price which the paying signer is willing to accept. The maximum `BASE_PRICE` they are willing to accept is compted as `MAX_GAS_PRICE - PRICE_PREMIUM`.
```
ashok-cl, line 52: `they are willing to accept is computed as` - typo

text/0007-gas-price-adjustment.md
```diff
@@ -0,0 +1,135 @@
...
+
+As far as the author is informed, there has been no precedent to this type of price regulation, other than EIP-1559 which has not been used in production yet.
+
+## Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+- Would a validator cartel emerge and follow price manipulation strategies for prolonged periods, such as fabricating demand, or proposing empty blocks?
+- How successful would the algorithm be at absorbing price volatility throughout the day? Would there be much need for price premiums?
+- Would the algorithm render the platform too expensive for "normal" use, or too inefficient (too much empty block space due to high price)?
```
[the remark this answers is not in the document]
osolmaz: What do you mean by 'low, medium and high' pricing? We are setting a price floor, and expect that users won't have to go above it most of the time.That contradicts different levels of pricing.

text/0007-gas-price-adjustment.md
```diff
@@ -0,0 +1,135 @@
...
+    BASE_PRICE = INITIAL_BASE_PRICE
+```
+
+This is to increase the speed of recovery from a catastrophic failure, by enabling as many users as possible to transact. We choose the value of `INITIAL_BASE_PRICE` low enough, so that resetting practically results in a floating price regime. That is, until the platform recovers from the failure and the algorithm catches up with the market price.
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+- The argument for this proposal favors *price stability* over *market efficiency*. In other words, we achieve stability by setting a price floor, and making users occasionally *pay higher* than the actual market price. If this argument is wrong, then the algorithm would be rendered useless.
```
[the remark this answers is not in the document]
osolmaz: Also, have you seen this? https://www.coindesk.com/ethereums-fees-mean-choosing-between-a-world-computer-and-a-financial-network

> In fee-bearing blockchains like Bitcoin, marginal, spammy, or non-monetary usage simply gets priced out over time.

text/0007-gas-price-adjustment.md
osolmaz: Other dapps? I don't understand the question, but I can try to guess.

If there are 2 dapps, say A and B, where users of A have much higher purchasing power than users of B. (e.g. A = Compound, B = an Ethereum game)

Then if gas price were to increase due to sudden increase in demand by A's users, B's users would be pushed out of the market. From the [same article](https://www.coindesk.com/ethereums-fees-mean-choosing-between-a-world-computer-and-a-financial-network):

> Today, a user sending a multi-million-dollar Tether transaction will most likely outbid someone deploying an Aragon DAO or minting an NFT.

text/0007-gas-price-adjustment.md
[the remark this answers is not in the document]
MParlikar: If we will set a price floor, that won't increase, this makes me happy.

text/0007-gas-price-adjustment.md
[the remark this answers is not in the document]
MParlikar: I"m with Vitalik on this one.  While I don't think fees for simple contracts should approach 'dollars' - I do agree that the high trust coming from blockchain should cost - commensurate with the level of trust it brings.

text/0007-gas-price-adjustment.md
MParlikar: Okay got it.  The issue really is around how full blocks are.

text/0007-gas-price-adjustment.md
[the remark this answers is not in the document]
osolmaz: But it's not about "should". There is no way around expensive prices once the market is ready to pay for it.

text/0007-gas-price-adjustment.md
```diff
@@ -0,0 +1,135 @@
...
+Then, a system transaction can be executed in any block proposed after `(n+1)*ADJUSTMENT_PERIOD`, which updates `BASE_PRICE` according to:
+
+```python
+if FULLNESS > TARGET_FULLNESS:
+    new_base_price = BASE_PRICE * (DENOMINATOR + ADJUSTMENT_RATE) / DENOMINATOR
+else:
+    new_base_price = BASE_PRICE * DENOMINATOR / (DENOMINATOR + ADJUSTMENT_RATE)
+
+if new_base_price > INITIAL_BASE_PRICE:
+    BASE_PRICE = new_base_price
```
afck, line 75: Or maybe:
```python
BASE_PRICE = max(new_base_price, INITIAL_BASE_PRICE)
```
Because if it's lower, then at least we can go back to `INITIAL_BASE_PRICE`?

afck (commented):
To be honest, I'm still skeptical about this approach in general, because as a user it doesn't help me much if I know:
> In the next 7 days, I will never pay less than 3$ to use the subway, but I might have to pay 1000$, and I might be unable to afford it at all.

It would be much more useful to know:
> In the next 7 days, I am guaranteed a seat in the subway for 5$ per day.

This proposal seems to guarantee a _minimum_ price, and cause many blocks to end up almost empty, preventing low-cost transactions from happening that would otherwise take place without displacing more important ones. What I'd want as a user is to know a _maximum_ price and have some guaranteed level of service.

I remember that as an alternative, something like "gas futures" came up a few months ago. Intuitively, these make much more sense to me: I would be able to buy a guaranteed amount of gas per block for the next week (day? month?) in advance, and no matter what happens to the market, I would be able to use them and guarantee that my dapp keeps working or my payments arrive.

I could still sell the gas futures instead, and make a profit, if I end up not needing them (or the price goes so high that I don't care about my dapp anymore :money_mouth_face:), but at least I know exactly what my costs will be for a guaranteed (except in case of equivocations/network outage/etc.) level of service.

text/0007-gas-price-adjustment.md
osolmaz: `BASE_PRICE` should not be able to go lower than `INITIAL_BASE_PRICE`. Did I make a mistake with the pseudocode?

afck: No, but the pseudocode would leave the `BASE_PRICE` at `105` if `INITIAL_BASE_PRICE` is `100` and `new_base_price` is `90`. I think setting it to `100` would make more sense.

Also, it's a strict inequality, so even if `new_base_price` is `100` and `INITIAL_BASE_PRICE` is `100`, it would leave `BASE_PRICE` at `105` instead of setting it to `100`.

osolmaz:
> This proposal seems to guarantee a minimum price, and cause many blocks to end up almost empty, preventing low-cost transactions from happening that would otherwise take place without displacing more important ones.

Yes. That is intentional. See the Drawbacks section.

> What I'd want as a user is to know a maximum price and have some guaranteed level of service.

An SLA which is satisfied 99.999% of the time is not possible without scalability. That was the point of introducing this mechanism. Before, there were no SLA. Now, we'll put forward an SLA, and optimistically say it will be satisfied 90% of the time.

> I remember that as an alternative, something like "gas futures" came up a few months ago. Intuitively, these make much more sense to me: I would be able to buy a guaranteed amount of gas per block for the next week (day? month?) in advance, and no matter what happens to the market, I would be able to use them and guarantee that my dapp keeps working or my payments arrive

Yes.

1. Gas futures are too complex for us to ship with mainnet.
2. Gas futures would likely have convoluted UX.
3. Gas futures don't solve the issue of gas being too expensive.

This proposal is much simpler than gas futures, and do not require a decision from the user. Otherwise, I agree that they would have complemented the market.

text/0007-gas-price-adjustment.md
osolmaz: I realize it now, sorry. Fixed

afck (approved):
OK, thanks for the clarification.

I'll approve this then: I don't think there's anything technically wrong with it, and I'm fine with it if that's the desired behavior. I'm just skeptical whether it's ideal from a user perspective.

afck:
I totally agree that gas futures are probably too complex for mainnet! The following are just some thoughts about the longer term.

> An SLA which is satisfied 99.999% of the time is not possible without scalability.

I'd still argue that it _is_ possible, in some sense, at least, say, 99.9%, which is probably _much_ more than this CEP can offer. It should be possible for the network to have 99.9% uptime, if validators do their job. And if it does, and you have gas futures up to one month into the future, then if you own such futures for the next month, you _are_ pretty much guaranteed that you can use them.

The model I have in mind is gas futures that apply to a particular block: So at block 1000, you can buy 50000 gas in block 1500. Then only the user who owns that particular future (you might sell it to someone else in the meantime, which has to happen on-chain) can submit a deploy for that particular block, and if nobody does, that part of the block must remain free. (Otherwise the validator has an incentive to ignore the deploy and replace it with one that pays more.)

So you'd not only buy a particular amount of gas, you'd buy a steady amount of gas in each of the next week's blocks (possibly for very different prices, depending on what time of the day that block is projected to happen in). If you don't use it, it remains unused.
In the end, I think that would let _less_ gas (i.e. space in block, idle CPU and bandwidth, etc.) go to waste than this CEP.

> Now, we'll put forward an SLA, and optimistically say it will be satisfied 90% of the time.

But isn't this a pretty bad kind of SLA? Let's look at two scenarios:

## Scenario 1

There's a huge price spike every day between 10:00 and 11:00, because during that period, my popular gambling dapp makes its computations and payouts. Otherwise the network is pretty idle. That means the price will be reduced until (either the initial price is reached or) it's so low that other, much less important transactions fill up roughly half of the blocks up to the target fullness.
Between 10:00 and 11:00, however, I outbid them by a large margin and fill up all blocks by myself.

Users either have to outbid me (no price stability, might as well not have this mechanism), or they make use of our "stable" regular price and basically have all of their transactions and dapps stalled between 10:00 and 11:00.

With 1-month gas futures, however, they could bid for the different blocks ahead of time, and essentially negotiate with me (via the free market) how important this 10:00-11:00 gas is for me. They would pay less in the blocks outside of that time window, and probably all of the blocks would get filled up pretty much 100%, because there's no minmum price. And within the 10:00-11:00 window, they would still have to outbid me, but could know well in advance whether they succeeded. If they didn't, or can't afford it, they can announce to their users that their dapp will have an outage, and they can make their payments outside that time window.

## Scenario 2

My gambling dapp now runs continuously for 13 hours a day and uses 80% of the blocks. During that period, the remaining 20% are easily filled with user transactions, so the median fullness is 100%. The base price rises until that isn't the case anymore, i.e. until the other users and dapps can't afford anything anymore. For the remaining 11 hours of the day, the blocks are basically empty, because the price is too high, even though there would be a lot of capacity.

In fact, if the target fullness is below 80% the price will rise until even _I_ can't afford running my dapp anymore, even though the chain has plenty of free capacity!

Again, with gas futures, people could buy those 11 hours worth of gas ahead of time for a much lower price, as well as those 20% during my dapp's activity.

afck:
> Gas futures don't solve the issue of gas being too expensive.

I'd argue that this CEP doesn't solve the issue either, but makes it worse. (See examples above.)

osolmaz:
> The model I have in mind is gas futures that apply to a particular block: So at block 1000, you can buy 50000 gas in block 1500. Then only the user who owns that particular future (you might sell it to someone else in the meantime, which has to happen on-chain) can submit a deploy for that particular block, and if nobody does, that part of the block must remain free. (Otherwise the validator has an incentive to ignore the deploy and replace it with one that pays more.)

We had followed this line of reasoning before, e.g. [here](https://hackmd.io/@onur/gas_reservation), and here are the problems we encountered:

- The UX of reserving block space on a single block is not that great. I'm guessing that users would be more comfortable with a futures contract where a price and inclusion is guaranteed for a given *day*. Because users may not exactly know the point in time they will need to transact. The problem is, relaxed conditions like these are not enough to have a working system.
- "Futures contracts" by definition require law enforcement, so that parties that don't honor the agreement are punished. But that is not possible in blockchains, so the only way we can encourage validators to honor "gas futures" agreements is through incentives. However, by doing that, we already lose some guarantees that may prove the idea impractical.

osolmaz:
> I'd argue that this CEP doesn't solve the issue either, but makes it worse.

I agree with your point. It keeps the price stable and artificially high, but pushes users with lower purchasing power out of the market.

I'm still open to reconsidering this.

afck:
Yes, I agree, to make futures work, they would _have_ to be for a specific block, and the UX would definitely be more complicated.

Not sure about the law enforcement: The rules of the blockchain can enforce that if a future for a particular place in a block has been sold, that space can only be used for a transaction using that future (i.e. signed by the current owner of that future). Validators could still decide to leave it empty and censor the transaction, of course, but I'd say the price the original buyer paid for the future would go to the validators who end up including the transaction, not to the ones including the purchase. (Those would have received the transaction fees for the purchase transaction, of course.) So the validators would have an incentive to include the transaction, otherwise the purchase price of the future would effectively get burned (or refunded? partially?) instead of paid out to them.

Anyway, sorry for going off-topic!

osolmaz:
> The rules of the blockchain can enforce that if a future for a particular place in a block has been sold, that space can only be used for a transaction using that future (i.e. signed by the current owner of that future).

You are right. For some reason, I was ignoring that fact. A rational actor would include that transaction, instead of leaving the reserved block space empty. Then it is indeed a viable solution.

If we can figure out the UX, I think this would be a better solution to ship to mainnet.

afck:
I'd love that! It sounds preferable to me.
But I didn't mean to derail all the plans. As I said, I'm not against this CEP, especially as an intermediate solution, if it's much simpler to implement.

text/0007-gas-price-adjustment.md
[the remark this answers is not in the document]
AlexanderLimonov: Well, no, the price floor is moving, with premia/bids on top.

text/0007-gas-price-adjustment.md
```diff
@@ -0,0 +1,134 @@
...
+
+CEP PR: [casperlabs/ceps#0007](https://github.com/casperlabs/ceps/pull/0007)
+
+Gas price is regulated algorithmically. A price floor exists and adjusts daily, based on fullnesses of blocks from the previous day, in order to absorb volatility and reduce congestion.
+
+## Motivation
+
+[motivation]: #motivation
+
+Gas prices are known to be subject to high volatility both in the short term and the long term. This is due to a combination of factors such as inelastic supply, [daily demand cycle](https://solmaz.io/2019/10/21/gas-price-fee-volatility/) and shifting trends. CasperLabs sees these issues with volatility as a key barrier to the adoption of public blockchain infrastructure by businesses. It is also recognized that price stability alone is not enough, and scalability is also a core concern in this arena.
```
AlexanderLimonov, line 15: We should probably find a way to somehow attach external documents to CEPs, or else mostly paste such things into the body itself, because if understanding the CEP requires the external information and your blog goes down, it's no good!

text/0007-gas-price-adjustment.md
```diff
@@ -0,0 +1,134 @@
...
+- `INITIAL_BASE_PRICE`: The value of `BASE_PRICE` at genesis. Also the *minimum value* for `BASE_PRICE`. Value: TBD.
+- `ADJUSTMENT_RATE`: The rate at which `BASE_PRICE` can move up or down per day. Value: `10_000_000_000` (1%).
+- `DENOMINATOR`: Fractions will be represented as trillionths of 1. Value: `1_000_000_000_000`.
+- `BLOCK_GAS_LIMIT`: Maximum gas allowed in a block. Value: TBD.
+- `TARGET_FULLNESS`: The value which will determine whether `BASE_PRICE` should move up or down. Value: `650_000_000_000` (65%).
+- `MIN_FINALIZED_BLOCKS`: The minimum number of blocks that must be finalized in an adjustment period. Otherwise, a fail-safe is triggered. Value: TBD.
+
+### Transaction Fields
+
+- `PRICE_PREMIUM`: The price premium specified for a given transaction.
```
AlexanderLimonov, line 51: Yeah, so, this is the part that I don't like from EIP-1559 and was, originally, thinking that we dropped it. The problem with a day-by-day plodding price discovery is that, _behaviorally_, it makes not an iota of difference to the users whether their money is going to the validators or poofing into non-existence. If the base price isn't matching what the people who, on the Ethereum platform, would have paid while generating a price peak, _we will still get the same price peak_. It's just that some of the tokens may be burnt (actually, are we proposing to do that, or not? Can't find it in the CEP, and that should be addressed). So we will have constructed this whole complicated machine for nothing.

My strong preference is for getting rid of bids altogether, making the price adjust quicker (every few eras perhaps), and hoping that we can implement genuine gas reservation before someone makes the first usable accelerator (given low use for the first two year, I think this is a realistic expectation).

AlexanderLimonov:
>   1. Gas futures are too complex for us to ship with mainnet.
>   2. Gas futures would likely have convoluted UX.
>   3. Gas futures don't solve the issue of gas being too expensive.

1. Well, that much is true.

2. Most users of gas futures would be developers of large dApps, not randos trying to run tutorial contracts, they can handle a bit of a UX complication if they're running a serious operation. Also, I don't think the UX is really the main problem with the gas futures. The main problem is that it's completely unclear how one would go about forcing or incentivizing validators to include transactions run on pre-paid gas, if there are competing alternatives willing to pay higher (through an accelerator, potentially).

3. None of this solves this issue. The point of having this, or the gas futures, is to provide a means for dApp operators to forecast their costs, whether they are high or low. What we're trying to mitigate here is disutility of risk, not to magically solve the scalability problem. Gas futures are going to be better for that than semi-fixed prices. However, as I argued before, I don't think this mechanism can achieve anything different than what is seen in Ethereum.

Also, I think the CEP needs to specify that it's adjusting the fiat price, which could (and, as planned, would) be recalculated into CASPER using Chainlink reports.

text/0007-gas-price-adjustment.md
osolmaz: There is web archive, but apparently the formulas aren't rendered correctly.
https://web.archive.org/web/20200828124936/https://solmaz.io/2019/02/24/scalable-reward-changing/

osolmaz: Link to web archive maybe?
https://web.archive.org/web/20200828125115/https://solmaz.io/2019/10/21/gas-price-fee-volatility/

osolmaz:
Closed, as it was not approved.

PR #8, opened 2020-08-25
Replace use of blake2 hash with blake3

ledgerleap:
[Rendered](https://github.com/ledgerleap/ceps/blob/master/text/0000-blake3-hashing-for-rust-node.md)


marc-casperlabs:
I have updated the PR slightly with some cleanups, which should make it ready for review. However I have a curious issue where GitHub will not let me make a PR against https://github.com/ledgerleap/ceps. This stems mainly from the fact that my local "fork" of the `ceps` repo predates the official one.

Sadly, I cannot change this (for example by deleting the repository and forking it again), as I have open PRs agains the official `ceps` that depend on said repository.

Of course, @ledgerleap is free to take the branch and manually merge it into theirs: https://github.com/marc-casperlabs/ceps/tree/blake3-hash

I am sorry for the inconvenience. =(

Fraser999 (approved):
This seems worthwhile to me.  The single downside I can think of is that BLAKE3 is relatively new and hence not as "battle tested".  However, given that it's not brand new, and that its designers are mostly the same as those for BLAKE2, and that the Rust implementation appears to have been written by the lead designer, I'd be happy to make the change.

EdHastingsCasperLabs:
I brought this up five or six months ago and discussed it w/ Henry & Michael. At the time we decided to wait a bit to see if blake3 would be widely adopted. I am for making the change in the abstract, but we have higher priority work that must get done before we put cycles towards it.

mpapierski:
I've did small experiment to see how big are the global state tries we hash with blake2 and if we could potentially benefit from the increased speeds. It looks optimistic so I'm approving it. I'm curious about possible performance gains in our transaction per second benchmarks.

mpapierski (approved):

marc-casperlabs (changes requested):
I am in general not opposed to this, but I have to insist on the fixes from https://github.com/marc-casperlabs/ceps/tree/blake3-hash being considered (see my comment). Right now the title of the document is just "Title", for example.

zie1ony (approved):
Rust, Python and JS ecosystems seams to have mature implementations of Blake3, so there is no technical obstacles here. Blake3 has Electric Coin Company (ZCash) behind it and it's a brand I trust.
I think it would be great if we could join the general effort of testing Blake3 in prod :)

marc-casperlabs:
I can finally open PRs again, thanks GitHub support :)

Here's my suggested changes, at least the file renaming part is required: https://github.com/ledgerleap/ceps/pull/1

Afterwards I will happily approve and merge it =)

momipsl:
My only hesitation with respect to adopting blake3b is on the client side, in particular JS & to a lesser degree Python.  If these language bindings are considered mature, i.e. audited & performant, then I would echo the notion to adopt blake3b.  I would also incidentally propose to adopt sr25519 as an ECC type.

xcthulhu:

PR #9, opened 2020-08-27
Client API

Fraser999:
[Rendered](https://github.com/Fraser999/ceps/blob/client-api/text/0009-client-api.md)

### DIFF
--- a/text/0009-client-api.md
+++ b/text/0009-client-api.md
@@ -82,17 +82,13 @@ using `ToBytes`, then hex-encoded.
 The path for the event stream is `/events`.  It will send all events to all subscribers from the
 moment they subscribe (i.e. no historical events will be sent).
 
-the available event types are
-* `block_finalized`
-* `block_added`
-* `deploy_finalized`
-* `deploy_processed`
-* `deploy_added`
+| Event            | Triggered by                                                           | Includes                                                  |
+|----------------- |----------------------------------------------------------------------- |---------------------------------------------------------- |
+| block_finalized  | `EraSupervisor` when creating a new `FinalizedBlock`                   | all included `DeployHash`es                               |
+| block_added      | `LinearChain` once the block has been stored                           | `BlockHash` and `BlockHeader`                             |
+| deploy_processed | `BlockExecutor` after each deploy is executed and transforms committed | `DeployHash`, all execution results and the commit result |
 
-The events corresponding to these subscriptions are as close as possible to those currently defined
-in [info.proto](https://github.com/CasperLabs/CasperLabs/blob/cd2e80286d4172c9d2e73a2c1f1271b2961a7527/protobuf/io/casperlabs/casper/consensus/info.proto#L91-L161).
-
-The first message returned will be the API version.  All subsequent ones will be JSON-encoded events.
+The first message sent by the server to any new subscriber will be the API version.  All subsequent ones will be JSON-encoded events.
 
 ### Use case
 
@@ -151,6 +147,7 @@ proposed API here does not require persisting extraneous data purely to satisfy 
 that data.
 
 There are also several other blockchain projects offering similar APIs, for example [Polkadot](https://polkadot.js.org/api/substrate/rpc.html).
+
 
 ## Unresolved questions
 
--- a/text/0009-client-api.md
+++ b/text/0009-client-api.md
@@ -4,7 +4,7 @@
 
 [summary]: #summary
 
-CEP PR: [casperlabs/ceps#0000](https://github.com/casperlabs/ceps/pull/0000)
+CEP PR: [casperlabs/ceps#0009](https://github.com/casperlabs/ceps/pull/9)
 
 This describes the client-facing API of the Casper Node.  It will be used by all clients interacting
 with a Casper Node, including the Rust client, the Python client and Clarity.
--- a/text/0009-client-api.md
+++ b/text/0009-client-api.md
@@ -1,6 +1,6 @@
 # Title
 
-## Client API of Casper Node
+## Casper Node: Client Facing API
 
 [summary]: #summary
 
@@ -34,6 +34,12 @@ While this may complicate things somewhat for clients, perhaps especially for Cl
 node software to be focused on its core responsibilities: accepting deploys, reaching consensus on
 their order, executing them and persisting changes to global state and the blockchain.
 
+Until we are able to provision some form of middleware which will be able to subscribe to nodes'
+event streams, and to which clients can direct their queries for metadata around deploy execution,
+the node will retain limited information in the deploy store as an interim measure.
+
+This information will be held as metadata associated with a given deploy and will include the cost
+of execution along with execution results.
 
 ## Reference-level explanation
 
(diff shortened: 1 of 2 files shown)


zie1ony:
1. Before we agree on such a "light" node we should define how the full CLarity intergration would look like.
2. If `GET /deploy` can't return the current deploy status, then making a deploy would look like:
    1. Subscribe to all of the deploy-related events at `/event` and maintain the connection.
    2. Make a deploy at `PUT /deploy`.
    3. Wait for events.
    
    While in general that shouldn't be a problem, some network connectivity issues might prevent a client from receiving deploy status. What the client should do in such a situation?

momipsl (commented):
0.  Before making any further comments it is worthwhile pointing out, as a previous reviewer has also stated, that this proposal is made purely from the perspective of a node stripped back to the minimal functional set required for correctness in the present tense.  The wider array of use cases that DApp's require is considered secondary, i.e. it is assumed that some form of off-chain worker process will build up and expose the deeper state lake required by typical DApp infrastructure - e.g. a read-only node.  This reviewer believes that correct minimal client facing API surface is one that supports a DApp development life-cycle out of the box.  Having said that, it is no doubt true that in production mode a validator operator will apply a custom ACL, i.e. endpoint whitelisting.

1.  See [api.pdf](https://github.com/CasperLabs/ceps/files/5147231/api.pdf) for this reviewer's notes regarding client facing APIs.

2.  Change title to 'Casper Node: Client Facing API'.

3.  JSON-RPC is a compromise between the richness of gRPC, the simplicity of (pseudo) REST and the composability of GraphQL.  Tooling chains will support the auto-generation of client proxies and the API becomes self-documenting.  It is widely used across the blockchain space.   I see no reason to use both JSON-RPC **and** SSE, JSON-RPC can be served over both HTTPS and WSS, thus event subscriptions can also be tunnelled through JSON-RPC.  

4.  The list of JSON-RPC endpoints are partly documented as if they were REST endpoints - e.g. /deploy "put", /deploy "get".  I prefer the following naming convention:  {scope}-{verb}-{noun} whereby scope is a grouping mechanism and verb is [get | set | subscribe | unsubscribe].  E.G. chain_get_block, account_set_deploy, state_get_value. 

5.  Events stream payloads should be minimal, i.e. rather than data they should return identifiers that can be resolved to data.  They should be grouped under the relevant scope to which they apply.  I definitely do not like the idea of an **all** event.   

6.    Re Prior Art.  There's loads of it.  Many other blockchains are using JSON-RPC, e.g. [Polkadot API](https://polkadot.js.org/api/substrate/rpc.html).

7.  Re Future Possibilities.  The Kafka solution proposed by Akosh with significant input by this review, is really an example of what could be done via an off-chain worker.  I.E. the node is running and well defined integration hooks, an off-chain worker is in turn pushing events to a Kafka cluster.  Further up-stream another API built onto of the Kafka cluster is being exposed to clients.

text/0009-client-api.md
```diff
@@ -0,0 +1,154 @@
...
+
+
+## Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+* Plain JSON requests and responses via a REST API is a viable alternative to JSON-RPCs.  There
+appears to be little difference in usability or implementation effort, although other projects in
+the blockchain space use JSON-RPC, so perhaps there would be some slight benefit for us to follow
+suit.
```
EdHastingsCasperLabs, line 116: I agree with this statement; I think the primary reason to use json-rpc over simple rest is to adhere to expected convention within the space.

text/0009-client-api.md
```diff
@@ -0,0 +1,154 @@
...
+
+* Websockets could be used instead of SSEs.  These should work for us, but they provide full-duplex
+connections, allowing back and forth messaging between the client and server.  Our use case is
+simpler: the client specifies what it wishes to subscribe to in the GET request, and all further
+traffic is from the node to the client.  As such, SSEs seems to be the simpler, better choice.
+
+* Multiple event streams could be provided by grouping events into categories.  It doesn't seem like
+this would be of any benefit, and if it results in a large increase of HTTP connections, could
+instead be detrimental.
```
EdHastingsCasperLabs, line 125: My inclination would be to do the simplest thing that works; if one event stream is sufficient I see no compelling reason to offer multiple streams until such a time as we have a concrete use case that demands it.

text/0009-client-api.md
```diff
@@ -0,0 +1,154 @@
...
+
+* Node could persist all ancillary information similarly to what is done in the Scala codebase.
+This would increase the complexity of node and decrease the performance.  It would also consume
+memory and disk resources.  It seems like a better solution to offload that to a separate process
+designed specifically to handle that information.
```
EdHastingsCasperLabs, line 130: My strong preference is to offload as much data as possible to some manner of middleware event stream consumer / data aggregator.

Having said that, we may need to store some kind of executed deploy summary so that we can answer basic questions about deploy disposition, as a short term measure until such a time as we can offer a middleware implementation.

text/0009-client-api.md
```diff
@@ -0,0 +1,154 @@
...
+
+In this way, the client's `put-deploy` and `send-deploy` subcommands could accept arguments specifying that the client
+wait until the deploy has transitioned to certain state.  Internally, the client will consume events (possibly reporting
+each) until the requested deploy state has been reached, at which point the client can disconnect.
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+* Clients will not be able to receive as rich information in response to requests for information as they currently do.
```
EdHastingsCasperLabs, line 106: I consider this to be more of a trade off than a draw back.

text/0009-client-api.md
```diff
@@ -0,0 +1,154 @@
...
+A shift in paradigm from the Scala node is that the node will not retain state relating to deploys
+or blocks other than what is required to satisfy the correct operation of the node.  In other words,
+the node will not retain state purely to be able to satisfy client requests to retrieve such
+ancillary state.  Rather the node will present an event stream to which clients may subscribe,
+allowing them to build up the information they require.
+
+So, for example, rather than being able to return information like the cost of executing a deploy,
+or an error message resulting from failed execution, this information will be announced by the
+relevant component (the `BlockExecutor` in this case) and emitted via SSEs by the `SseServer`
+component.
```
AbnerZheng, line 31: Does the `BlockExecutor` support query by `deployHash` or clients could only listen to the event stream?

There are some use cases that users want to know the cost or status of deploy that they deployed some days ago.

EdHastingsCasperLabs: Currently, we don't keep track of deploy status directly in node-rs. The comment I added above "...we may need to store some kind of executed deploy summary so that we can answer basic questions about deploy disposition, as a short term measure until such a time as we can offer a middleware implementation..." acknowledges that we might need to add some minimal state in the short to mid term until we have put more effort into the event stream concept.

AbnerZheng: Thank you for your explanation. If `node-rs` could answer basic questions about deploy disposition, I think the client API is sufficient for Clarity.

Fraser999:
@zie1ony 
> Before we agree on such a "light" node we should define how the full CLarity intergration would look like.

It seems like Abner is comfortable with the suggested API in terms of CLarity, given that the original proposal has been updated to support some limited Deploy metadata (gas cost/exec results) to be held by the node and returned in relevant RPCs.

> While in general that shouldn't be a problem, some network connectivity issues might prevent a client from receiving deploy status. What the client should do in such a situation?

I think this is also answered by the additional metadata being provided as above.

Fraser999:
@momipsl 
> Before making any further comments it is worthwhile pointing out, as a previous reviewer has also stated, that this proposal is made purely from the perspective of a node stripped back to the minimal functional set required for correctness in the present tense. The wider array of use cases that DApp's require is considered secondary, i.e. it is assumed that some form of off-chain worker process will build up and expose the deeper state lake required by typical DApp infrastructure - e.g. a read-only node. This reviewer believes that correct minimal client facing API surface is one that supports a DApp development life-cycle out of the box. Having said that, it is no doubt true that in production mode a validator operator will apply a custom ACL, i.e. endpoint whitelisting.

I believe this is at least partly addressed in the addition of some limited Deploy metadata being held by the node and returned in relevant RPCs.  There is still a desire to keep the node's responsibilities minimal, and the CEP reflects that.  However, should we decide in the future that the node _should_ take on further duties to avoid the requirement of separate data aggregators, then I don't believe what is proposed here will present a barrier to that effort.  Such a refactoring would likely comprise mainly additions to the codebase rather than reworking or deleting code.

> Change title to 'Casper Node: Client Facing API'.

Done

> JSON-RPC is a compromise between the richness of gRPC, the simplicity of (pseudo) REST and the composability of GraphQL. Tooling chains will support the auto-generation of client proxies and the API becomes self-documenting. It is widely used across the blockchain space. I see no reason to use both JSON-RPC and SSE, JSON-RPC can be served over both HTTPS and WSS, thus event subscriptions can also be tunnelled through JSON-RPC.

I considered WSS and listed it in the "Rationale and alternatives" section.  WSS seems like overkill for our needs since the client, once subscribed, doesn't need to send any further messages to the server.  Also, regardless of whether we choose Websockets or SSEs, I don't believe JSON-RPC suits being used in an event stream context.  It implies a single request to single response flow as opposed to single request, stream of responses.

> Furthermore binary data can be transmitted over JSON-RPC whereas SSE only supports text - however admittedly this might be irrelevant this system.

I'm not sure that's correct.  JSON-RPC uses JSON for encoding - we'd need to hex-encode (or base64-encode or similar) binary data for use in a JSON-RPC.

> The list of JSON-RPC endpoints are partly documented as if they were REST endpoints - e.g. /deploy "put", /deploy "get". I prefer the following naming convention: {scope}-{verb}-{noun} whereby scope is a grouping mechanism and verb is [get | set | subscribe | unsubscribe]. This convention will allow the API more room to breathe overtime. E.G. chain_get_block, account_set_deploy, state_get_value.

Updated in the doc now - thanks.

> Events stream payloads should be minimal, i.e. rather than data they should return identifiers that can be resolved to data. They should be grouped under the relevant scope to which they apply.

That's a good point about the size of events in the stream.  I updated the doc to give more info on the events and their triggers.  I also reduced the count to three, as I believe previously there was a degree of duplicate info being sent.

> I definitely do not like the idea of an all event.

By that, do you mean you don't like the idea of a client subscribing to receive all events the node produces?  The CEP has been changed to remove filtering events at the server, so effectively clients only have the capacity to subscribe to all or no events.  I give the rationale in the final bullet point of "Rationale and alternatives" - does it seem reasonable?

> Re Prior Art. There's loads of it. Many other blockchains are using JSON-RPC, e.g. Polkadot API.

Good point - added now.

text/0009-client-api.md
Fraser999: The CEP has been updated to reflect this now.

text/0009-client-api.md
Fraser999: I rephrased the sentence slightly.

text/0009-client-api.md
Fraser999: Great - thanks!

AbnerZheng (approved):

PR #11, opened 2020-08-28
Storage token

osolmaz:
Users need to own *storage tokens* (STR) to store data in the global state.

[Rendered](https://github.com/osolmaz/ceps/blob/storage-token/text/0011-storage-token.md)


osolmaz:
The reference-level explanation requires low-level details (probably) involving details on how we implement our Patricia-Merkle tries, how we meter for storage, and charge transaction fees. The current state of the doc outlines the idea well enough to start a discussion. For low-level details, we can collaborate in the upcoming months.

afck (commented):
What's the advantage of a separate STR token compared to a fixed storage fee and refund? Doesn't it just introduce an unnecessary extra step?
I.e. if I
* buy 1 STR for 100 CASPER,
* make my transaction and store 1 byte,
* make another transaction and free my 1 STR again,
* burn my STR and get back 100 CASPER,

then why not just simplify the process and say that storing 1 byte costs 100 CASPER and freeing 1 byte yields a 100 CASPER refund?

Also, why is it necessary to track any ownership? If you call a smart contract and have to pay 100 CASPER because you cause the contract to store 1 byte, and then I call the same contract and cause that byte to be freed, why not give me the 100 CASPER refund, regardless of whether I own any STR tokens?

**On the other hand:**

I _do_ see a value in a separate token if we define a maximum global state size, if it works like emissions trading. We'd create one STR token for every byte we want to allow in the global state, and let users trade them, and determine the price. Whenever you free a byte, a new STR is minted for you, and whenever you store one byte, you must burn an STR.

The advantage is that we don't have to guess in advance what percentage of CASPER will end up being bound in storage.

With this CEP, if the total supply is (I'm using small numbers here for illustration), say, 10000 CASPER and we want a maximum global state size of 50 bytes, we'd have to guess: If we guess that 5% of tokens will be used for STR/storage, we'd set the price of 1 byte to 10 CASPER. If we guessed wrong, the price will be unnecessarily high and we'll use much less storage than our validators could provide, or we'll use _more_ storage than we had hoped, making syncing difficult and creating additional hardware cost for validators (or even failure).

If we don't set a fixed price however, we'd just say we'll only ever mint 50 STR tokens and let users trade with them. That way it's guaranteed that the global state can never exceed 50 bytes (well, plus our auction contract and stuff), and on the other hand the market will probably make sure that those 50 bytes get actually used by the users who need them most.

text/0011-storage-token.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Motivation
+
+[motivation]: #motivation
+
+The most straightforward way of pricing blockchain resources is to charge a fixed amount of gas per VM opcode, as a one-time payment. However, unlike computation which is used and freed immediately, persistent storage of data in the global state means that the resource continues to be used, unless it is freed manually by the user at a later time. For that reason, one-time payment is not the ideal model for pricing persistent storage.
+
+To remedy this issue, Ethereum overprices storage, which curbs the growth rate of the global state. Ethereum also implements a *gas refund* mechanism, where users can discount gas usage of their transactions, if they free up previously used space. However, this arrangement does not ensure an upper bound to global state size. It grows at a linear rate, and one can see that it will require full nodes to have higher and more expensive storage[^1]. Unless an alternative model is implemented, the requirements could increase indefinitely, decreasing the number of full nodes due to increased costs. This can have adverse affects on the network.
```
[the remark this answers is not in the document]
ashok-cl: @osolmaz Can you respond to @afck 's suggestion / comment above?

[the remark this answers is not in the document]
osolmaz: Yes, I plan to progress this CEP soon.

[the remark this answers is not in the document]
osolmaz: @afck I'm just curious, it seems like nothing is changed in this suggestion. Or maybe I am not seeing it :S

osolmaz:
> I do see a value in a separate token if we define a maximum global state size, if it works like emissions trading. We'd create one STR token for every byte we want to allow in the global state, and let users trade them, and determine the price. Whenever you free a byte, a new STR is minted for you, and whenever you store one byte, you must burn an STR.

@afck yes, that was my rationale for this proposal. The only advantage over the refund modal is having an upper bound on global state size. This is useful keeping it feasible to run a full node, and keep the network decentralized. 

I added an estimation of `STR_PRICE` for a certain desired upper bound. I also compared with current Ethereum levels. Finally, I added a question about how to handle STR price changes.

> Also, why is it necessary to track any ownership? If you call a smart contract and have to pay 100 CASPER because you cause the contract to store 1 byte, and then I call the same contract and cause that byte to be freed, why not give me the 100 CASPER refund, regardless of whether I own any STR tokens?

You are right. I got rid of that part.

text/0011-storage-token.md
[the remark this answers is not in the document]
afck: I had to re-read it twice, but it's `affects` → `effects`.

text/0011-storage-token.md
```diff
@@ -16,20 +16,34 @@ The most straightforward way of pricing blockchain resources is to charge a fixe
...
 
-Here, we propose an alternative model which leverages the scarcity of CASPER to limit the growth of global state, and increase the allocative efficiency of storage.
+Here, we propose an alternative model which leverages the scarcity of CSPR to limit the growth of global state, and increase the allocative efficiency of storage.
 
 ## Guide-level explanation
 
 [guide-level-explanation]: #guide-level-explanation
 
-We introduce the concept of *data ownership*: every bit of data stored in the global state is associated with an account. Accounts must hold storage tokens in order to store data, denoted internally by the symbol STR. STR is indivisible and is the smallest denomination of itself. 1 STR gives its holder the right to store 1 byte of data.
+Every bit of data stored in the global state is associated with an account. Accounts must hold storage tokens in order to store data, denoted internally by the symbol STR. STR is indivisible and is the smallest denomination of itself. 1 STR gives its holder the right to store 1 byte of data.
```
afck, line 25: `bit` → `byte`
But still: It doesn't need to be associated with an account for the purposes of this proposal, does it? I pay when I store it, and you get back the STR when you free the data.

afck (commented):
I'm not sure about enforcing the upper bound via an explicit (as opposed to determined by the market) `STR_PRICE` and the total CSPR supply, because we have no idea what percentage of CSPR will be used for storage. If it's 1%, only 1% of the maximum storage will be used, which would be a waste of available resources.

I'd rather have a fixed supply of CSPR that are traded freely, so that even if demand is low, the available storage actually gets used.

text/0011-storage-token.md
osolmaz: I updated the sentence again
> Every byte of data stored in the global state belongs to an account.
Does it sound OK now?

afck: I don't understand why the association between the byte and the account is relevant for this proposal.

osolmaz: I say to highlight the fact that there can be no data that is stored by the users without going through the STR minting process. If it still sounds irrelevant, I'll remove the sentence.

afck: Sure, it's probably fine to leav it like that.
It just reads a bit as if that association were still relevant after the initial store operation.

osolmaz: Removed that first sentence and added the point to the next paragraph. This should be resolved now.

xcthulhu:
Shelving this idea for now.

PR #14, opened 2020-09-15
Native currency name, symbol and denomination

osolmaz:
The Casper Network has a native currency, called Casper, and denoted by the ticker symbol CSPR.

[Rendered](https://github.com/osolmaz/CEPs/blob/cspr-denomination/text/0014-native-currency.md)

### DIFF
--- a/text/0014-native-currency.md
+++ b/text/0014-native-currency.md
@@ -4,7 +4,7 @@
 
 [summary]: #summary
 
-CEP PR: [casperlabs/ceps#0000](https://github.com/casperlabs/ceps/pull/0000)
+CEP PR: [casperlabs/ceps#0014](https://github.com/casperlabs/ceps/pull/0014)
 
 The Casper Network has a native currency, called Casper, and denoted by the ticker symbol CSPR.
 
--- a/text/0014-native-currency.md
+++ b/text/0014-native-currency.md
@@ -33,6 +33,14 @@ We represent all Casper related quantities in terms of motes at the implementati
 
 The initial supply of Casper at mainnet launch will be 10 billion. The issuance schedule is out of the scope of this proposal.
 
+### Ortography and Usage
+
+Casper, CSPR and mote are all singular and always used without the plurality suffix -s.
+
+- Casper is always to be used with capital C: 100 Casper.
+- CSPR is always to be used with all capital letters: 100 CSPR.
+- Mote is used as a common noun, so the first letter is capitalized only when necessary, like at the beginning of a sentence: 100 mote.
+
 ## Reference-level explanation
 
 [reference-level-explanation]: #reference-level-explanation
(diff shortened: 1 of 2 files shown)


piotr-dziubecki:
So if for Ethereum token we use ETH and its smallest unit 'wei' can we also officially define how we write mote/motes (I guess lower case and we don't need abbreviation?)

text/0014-native-currency.md
```diff
@@ -0,0 +1,94 @@
...
+
+We define `total_supply` as the total supply of a blockchain's native currency, and `decimals` as the number of decimal places representing its smallest subdenomination.
+
+Then supply of the base unit is computed as `base_supply = total_supply * 10^decimals`.
+
+| Blockchain | Total supply | Decimals | Base supply |
+|-|-|-|-|
+| Bitcoin | 21 million | 8 | ~10^15.32 |
+| Ethereum | ~100 million | 18 | ~10^26 |
+| Polkadot | ~1 billion | 10 | ~10^19 |
```
afck, line 71: Does Polkadot use 64-bit numbers for amounts? It looks like that's the intention, because
```python3
>>> 2**64 / 10**19
1.8446744073709551
```
With a 2% increase per year, this means our total supply would fit into a `u64` for 30 years:
```python3
>>> log(2**64 / 10**19) / log(1.02)
30.920280414900077
```
Each zero we remove would add about 116 years to that. But I'm comfortable with having our very own "year 2050 problem" and planning for a network upgrade within the next 30 years, too. :+1: 

If we want to use a `u128` from the beginning, however, it feels like we'd be wasting half of the bits. If we do that we might as well add more zeros and smaller rounding errors.

afck (approved):

osolmaz:
> So if for Ethereum token we use ETH and its smallest unit 'wei' can we also officially define how we write mote/motes (I guess lower case and we don't need abbreviation?)

Yes, this CEP is supposed to be that definition. Casper is written and pronounced as it is, but it can also be referred to with the symbol CSPR. Motes can be written lowercase, and used with SI prefixes for greater subdenominations.

text/0014-native-currency.md
osolmaz: Polkadot uses 128-bit too. It looks like it's the norm

https://github.com/paritytech/polkadot/blob/96af6eadd5649e7faef75ddfe3cc181bd1567155/core-primitives/src/lib.rs#L63

osolmaz:
@piotr-dziubecki I added a subsection about orthography and usage

AlexanderLimonov (approved):
We could probably use a somewhat more comprehensive discussion of the potential performance/memory/implementation issues arising from using very big integers types, but this is good otherwise and we can amend later by means of another CEP

piotr-dziubecki:
image
@abner could you check this comments from discord against this CEP

PR #15, opened 2020-09-21
Gas Payment Model

zie1ony:
CEP. that describes how users could pay gas.

[Rendered](https://github.com/zie1ony/ceps/blob/payments/text/0015-gas-payments.md)


mpapierski (commented):
Impact on the existing code and behavior is big obviously, but if we can get this right, we can definitely remove a lot of code and simplify the platform.

- If we're to remove payment code, wouldn't it make sense to add special `ActionThreshold` for paying. So only associated accounts greater than the `payment` thresholds could be used for paying for execution. Or would that be a "deploy" weight? 
- I'd put a removal of a payment code as one of primary goals of this. (but keep in mind we still need to "finalize" the gas spent)
- If a contract pays for itself do we charge exactly 0 gas for it, or is there some overhead (i.e. for `call_contract` if anything)?
- That being said I'm interested how would gas accounting could work in case of a contract doing `call_contract` on a contract that pays for itself?

text/0015-gas-payments.md
```diff
@@ -0,0 +1,90 @@
...
+Most of the systems built on top of smart contracting platforms use tokens 
+and it is reasonable to put a lot of effort in designing a good token standard 
+or discover gaps in core functionalities, that should be fixed before 
+the Mainnet. The most popular token standard right now is the ERC20. We have 
+implemented its Casper version here https://github.com/casperlabs/erc20.
+We think we can do better and offer our future community a token standard 
+that avoids the "approve and call" pitfall. 
+This work will be the subject of future CEP.
+
+#### CSPR Interface
```
AlexanderLimonov, line 20: Can you elaborate on what exactly is problematic about ETH not implementing a good token standard? What constitutes a good token standard? Is there an example among existing ERC standards that you would consider good?

text/0015-gas-payments.md
```diff
@@ -0,0 +1,90 @@
+# Casper Payments: Gas Payment Concepts
+
+## High level view
+This is the first of a few CEPs on Casper Payment model. 
+The motivation for this work is to make Casper much more user and dev 
+friendly than it is now. There are a few topics here, that mix. The change 
+to one of them might cause the domino effect on others, but we'll 
+try to chunk them into logical units.
+
+#### Token Standard
```
AlexanderLimonov, line 10: Maybe you could go into a bit of detail about the flexibilities and inflexibilities of current payment code, which, according to my prior understanding, gave us significantly more flexibility compared to ETH? Or was I wrong? I never really looked into it closely.

text/0015-gas-payments.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Gas Payments
+In this section, we'll cover ways of paying for deploys that were discussed
+in a past and we'll decide what we want to support.
+
+### Self Payment
+This is the simplest case. User should be able to use its own CSPRs to pay
+for the execution of its own deploy (WASM or contract call).
+
+### Delegeted Payment
```
AlexanderLimonov, line 46: This is a good start, but maybe there should be a whole CEP dedicated specifically to Delegated Payments? It's going to be a complex issue to describe in the full CEP format, which present document does not follow.

AlexanderLimonov (commented):
I would like to see more about delegated payments (in fact, maybe the CEP should just focus on that), or, if the scope cannot be constrained, a detailed look into the tradeoffs between implementing some relatively rigid standard and just having a free for all Turing-complete payment system. But, honestly, I think dealing with the latter should be its own separate CEP.

Additionally, this should be converted to standard CEP format, to force discussion of downsides, alternative choices, etc.

zie1ony:
> If we're to remove payment code, wouldn't it make sense to add special ActionThreshold for paying. So only associated accounts greater than the payment thresholds could be used for paying for execution. Or would that be a "deploy" weight?

I think it should be deploy weight.

zie1ony:
> * If a contract pays for itself do we charge exactly 0 gas for it, or is there some overhead (i.e. for call_contract if anything)?
> * That being said I'm interested how would gas accounting could work in case of a contract doing call_contract on a contract that pays for itself?

These questions are hard to answer and lead to very complex solutions. That's why I don't think we should pursue this feature.

text/0015-gas-payments.md
```diff
@@ -0,0 +1,90 @@
...
+### Royalty Fee
+Another idea that was considered was Royaliry Fee for smart contract owner. As an user Alice 
+would have to pay an additional fee for using a contract, that would go to the contract owner. 
+
+This feature is easy to implement on the smart contract layer inside the contract,
+so we see no need to add that feature to the system. Fees management inside
+contracts are rich pieces of logic and it would require a very complex solution 
+to avoid execution smart contracts for that part. 
+
+### Removing Payment Code
```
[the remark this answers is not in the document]
zie1ony: To me the deploy should contains:
* `session_from` - session code context,
* `session_signatures` - list of signatures of associated keys that validate `session_from`,
* `payment_from` - account that pays for the execution,
* `payment_signatures` - list of signatures of associated keys that validate `session_from`.

`payment_from` and `payment_signatures` could be empty and then `session_from` should be treated as payment context.

text/0015-gas-payments.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Delegeted Payment
+Alice and Bob should be able to execute the following scenario:
+
+1. Alice creates a Session Code (contract call) and signs it with her private key. 
+2. Alice sends it to Bob.
+3. Bob adds Payment Code information and signs it with his private key.
+4. Bob sends it to the network.
+5. Network executes the deploy in a way that:
+    a. Session Code executes as Alice (her AccountHash is returned when calling `get_caller`),
```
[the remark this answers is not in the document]
zie1ony: My previous comment answers that https://github.com/CasperLabs/ceps/pull/15#discussion_r496502637

text/0015-gas-payments.md
```diff
@@ -0,0 +1,90 @@
...
+This is the simplest case. User should be able to use its own CSPRs to pay
+for the execution of its own deploy (WASM or contract call).
+
+### Delegeted Payment
+Alice and Bob should be able to execute the following scenario:
+
+1. Alice creates a Session Code (contract call) and signs it with her private key. 
+2. Alice sends it to Bob.
+3. Bob adds Payment Code information and signs it with his private key.
+4. Bob sends it to the network.
```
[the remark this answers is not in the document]
zie1ony: This could work in two ways:
1. (proposed version), where Alice creates a session code with signatures and sends it to Bob. Then Bob adds payment and of course could send it back to Alice.
2. Bob creates a session code (sets contract hash + arguments + session_from), signes it a payer and sends to Alice for the final session code signature. Then it's up to Alice to send it to the network or back to Bob.

Our of those two, the first one I expect the first one to be used much more often, as it should be in the Bob's best interest to execute the transaction. On the other hand Bob would be in the best position to front-run Alice. What do you think?

text/0015-gas-payments.md
```diff
@@ -0,0 +1,90 @@
...
+That enables Alice to interact with the network without her having to obtain CSPR tokens for gas.
+
+### Contract Payment
+It was considered may times to have a feature that allows Alice to call 
+a contract that can pay for the execution without a need for her to have CSPR tokens.
+
+To have that, the contract would have to perform some amount of computations 
+to check if Alice is eligible for a free execution.
+
+Unfortunately, this scenario is a subject of a spam attack, where Alice could
```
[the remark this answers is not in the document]
zie1ony: Agree, but that's a discussion for another CEP.

text/0015-gas-payments.md
zie1ony: > Can you elaborate on what exactly is problematic about ETH not implementing a good token standard? 

Uniswap v1 is a good example. Because the Ether doesn't implement ERC20 interface they had to implement more exchange methods:
* Ether to ERC20
* ERC20 to Ether
* ERC20 to ERC20

If Ether would implement ERC20, then the last one would be enough. The solution is to put Ether inside the ERC20 - https://weth.io/ But that creates the situation, that some platforms expects WETH, because that was easier for them to write, and some expects plain Ether, because they optimize for gas cost (plain Ether transferts are cheaper).

> What constitutes a good token standard?

I think the adoption is the best indicator. That's why iterating on top of the ERC20 is a good idea.

>  Is there an example among existing ERC standards that you would consider good?

ERC20. I know, I'm boring :smile:

text/0015-gas-payments.md
zie1ony: A few months ago I have tested what's possible to achieve with payment codes. I tried to write a [contract](https://github.com/CasperLabs/contract-based-payment/blob/ECO-229-payment-code/contract/src/lib.rs), that pays for the execution. First problem was that it required user to have the minimal amount of tokens on in the main purse. The second thing was a problem with refunds. The system returns unused tokens to the caller, not to the purse that payed for the execution. Of course, it could be fixed and we could extend payment codes with more features, but that requires a good design that prevents spam attack, which we don't have at the moment.

My statement is that there is no way to write something different than the standard payment with a current implementation. I would love someone to prove me wrong.

text/0015-gas-payments.md
zie1ony: The high level proposition of delegated payment + removal of payment codes should be enough to catch obvious problems and estimate the work required. The detail spec should be done in cooperation with a core team member, that knows the code internals. If the CEP has enough support, I'll spend more time on it.

PR #17, opened 2020-09-23
Gas futures

osolmaz:
A certain percentage of future block space is reserved and auctioned off. Said block space can only be filled by transactions signed by gas futures holders.

[Rendered](https://github.com/osolmaz/ceps/blob/gas-futures/text/0017-gas-futures.md)

### DIFF
--- a/text/0017-gas-futures.md
+++ b/text/0017-gas-futures.md
@@ -4,7 +4,7 @@
 
 [summary]: #summary
 
-CEP PR: [casperlabs/ceps#0000](https://github.com/casperlabs/ceps/pull/0000)
+CEP PR: [casperlabs/ceps#0017](https://github.com/casperlabs/ceps/pull/0017)
 
 A certain percentage of future block space is reserved and auctioned off. Said block space can only be filled by transactions signed by gas futures holders.
 
--- a/text/0017-gas-futures.md
+++ b/text/0017-gas-futures.md
@@ -30,7 +30,7 @@ X <= BLOCK_GAS_LIMIT - FUTURES_GAS_LIMIT
 
 where `X` is the total gas usage of transactions signed by users that do not hold gas futures. Blocks that do not fulfill this condition are deemed invalid.
 
-Gas futures are designated for consecutive periods of fixed length. In this proposal, we will focus on daily gas futures, but the same instrument can be extended to designate weekly, monthly, and so on. Gas futures are represented by tokens which are auctioned off continuously, some fixed duration preceding the futures date. (In this proposal, "futures" and "futures tokens" are used interchangeably.) These tokens will be designated in the format `GASYYYYMMDD`, and will be sold off in an on-chain auction `AUCTION_PERIOD` days before the futures date. In this proposal, we will assume `AUCTION_PERIOD = 180`, i.e. 6 months.
+Gas futures are designated for consecutive periods of fixed length. In this proposal, we will focus on daily gas futures, but the same instrument can be extended to designate weekly, monthly, and so on. Gas futures are represented by tokens which are auctioned off continuously, some fixed duration preceding the futures date. (In this proposal, "futures" and "futures tokens" are used interchangeably.) These tokens will be designated in the format `GASYYYYMMDD`, and will be sold off in an on-chain auction `GAS_AUCTION_PERIOD` days before the futures date. In this proposal, we will assume `GAS_AUCTION_PERIOD = 180`, i.e. 6 months.
 
 **Example:** Holders of the token `GAS20200901` will be able to include transactions on September 1, 2020 at the price they have agreed on at the auction. The auction takes place exactly 180 days before the delivery date, for the whole day, on March 5, 2020.
 
@@ -51,7 +51,7 @@ FUTURES_GAS_LIMIT[i] = BLOCK_GAS_LIMIT * FUTURES_SOLD[i] * FUTURES_SIZE / GAS_PE
 
 ### Auction
 
-The auction for `GASYYYYMMDD` begins at tick `X - AUCTION_PERIOD * TICKS_IN_DAY` where `X` is the tick corresponding to `YYYY-MM-DD` 00:00 in UTC, and `TICKS_IN_DAY = 24*60*60*1000` is the number of ticks in one day. The auction can go on until the whole quota is sold off.
+The auction for `GASYYYYMMDD` begins at tick `X - GAS_AUCTION_PERIOD * TICKS_IN_DAY` where `X` is the tick corresponding to `YYYY-MM-DD` 00:00 in UTC, and `TICKS_IN_DAY = 24*60*60*1000` is the number of ticks in one day. The auction can go on until the whole quota is sold off.
 
 The auction price is determined by the validators collectively, who announce the price they are willing to accept. The futures price is computed by taking a weighted average:
 
--- a/text/0017-gas-futures.md
+++ b/text/0017-gas-futures.md
@@ -22,36 +22,47 @@ We propose gas futures to enable users to hedge themselves against future gas pr
 
 [reference-level-explanation]: #reference-level-explanation
 
-Prior to this proposal, each block had a `BLOCK_GAS_LIMIT` which determined the throughput of the blockchain. We introduce a protocol parameter `FUTURES_GAS_LIMIT <= BLOCK_GAS_LIMIT`. This portion of the block space will be exclusive to gas futures holders, that is
+Prior to this proposal, each block had a `BLOCK_GAS_LIMIT` which determined the throughput of the blockchain. We introduce a protocol parameter `MAX_FUTURES_GAS_LIMIT <= BLOCK_GAS_LIMIT`. This portion of the block space will be exclusive to gas futures holders, that is
 
 ```python
-X <= BLOCK_GAS_LIMIT - FUTURES_GAS_LIMIT
+X <= BLOCK_GAS_LIMIT - FUTURES_GAS_LIMIT[i]
 ```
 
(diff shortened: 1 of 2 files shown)


osolmaz:
I think the reference level explanation of gas futures is ready for an initial review. I basically followed through @afck 's idea [here](https://github.com/CasperLabs/ceps/pull/7#issuecomment-682023948). I will finalize the other sections soon.

Let me know if you see something that doesn't make sense.

text/0017-gas-futures.md
```diff
@@ -0,0 +1,147 @@
...
+
+Prior to this proposal, each block had a `BLOCK_GAS_LIMIT` which determined the throughput of the blockchain. We introduce a protocol parameter `FUTURES_GAS_LIMIT <= BLOCK_GAS_LIMIT`. This portion of the block space will be exclusive to gas futures holders, that is
+
+```python
+X <= BLOCK_GAS_LIMIT - FUTURES_GAS_LIMIT
+```
+
+where `X` is the total gas usage of transactions signed by users that do not hold gas futures. Blocks that do not fulfill this condition are deemed invalid.
+
+Gas futures are designated for consecutive periods of fixed length. In this proposal, we will focus on daily gas futures, but the same instrument can be extended to designate weekly, monthly, and so on. Gas futures are represented by tokens which are auctioned off continuously, some fixed duration preceding the futures date. (In this proposal, "futures" and "futures tokens" are used interchangeably.) These tokens will be designated in the format `GASYYYYMMDD`, and will be sold off in an on-chain auction `AUCTION_PERIOD` days before the futures date. In this proposal, we will assume `AUCTION_PERIOD = 180`, i.e. 6 months.
```
afck, line 33: Let's make sure we call it `GAS_AUCTION_PERIOD` in the code to disambiguate. There's also the validator slot auction (which even has its own `AUCTION_DELAY`).
For the CEP it's fine, of course.

afck (commented):
I imagine it would be more useful if each GFET weren't tied to a day but to a specific block at a specific height.
Otherwise the GFET holders for a particular day face the same problem this CEP is trying to address: They might all intend to use their futures at the same hour of the day and then won't be able to get their transactions included _despite_ holding a gas future.

text/0017-gas-futures.md
osolmaz: Done

osolmaz:
> Otherwise the GFET holders for a particular day face the same problem this CEP is trying to address: They might all intend to use their futures at the same hour of the day and then won't be able to get their transactions included despite holding a gas future.

I see a couple of problems (which we have discussed a little bit already):

- The block at a certain tick/height might end up not being proposed after all, due to large round exponents.
- High storage overhead, because you need to store block-by-block information instead of day-by-day.
- More complicated UX. So much so that it might be a deal-breaker. It is not possible to estimate which at which time of the day the users will need to transact. Once the day comes, it might be improbable for gas futures users to time their transactions at such an exact level.

But I acknowledge the problem with daily reservation. Daily demand cycle shows that futures holders do not submit any transactions for 12 hours, and suddenly submit more than what `FUTURES_GAS_LIMIT` can contain.

In general, the larger the window size, the better the UX. But large window sizes can cause contracts to be unfulfilled. Maybe this is a trade-off between uncertainty and window size. A window size of 1 round is IMO infeasible, but we could possibly work with 3, 2 or even 1-hour window sizes. This would allow say a dapp provider with majority of users in the US vs in China to choose which time of the day they would transact.

afck:
I still have trouble imagining that this could work in any way other than per-block.

> block at a certain tick/height might end up not being proposed

I'd make it per block _height_, not per tick: At every height, a block will eventually be proposed and finalized. It might just be in a different tick than expected. The closer the delivery height, the more likely it becomes that the block will be at a particular timestamp, which will be reflected in the price the futures are traded at.

> More complicated UX.

But if there are functions like "buy 5 gigagas in each of the next 200 blocks" it might not be that bad. And it would guarantee you that (as long as the network itself does) your dapp will run smoothly in every block.
Hourly futures just guarantee that your dapp will be able to act at some time within each hour, but you have no control over when exactly. Daily ones are even worse.

> High storage overhead

That might be true, yes. But it feels like it should be solvable. E.g. we could have a single gas future token that gives you "5 gigagas in each block with height 1000 to 1200", instead of 200 individual tokens, and provide ways to split it up if necessary.

AlexanderLimonov:
> 
> 
> I still have trouble imagining that this could work in any way other than per-block.
> 
> > block at a certain tick/height might end up not being proposed
> 
> I'd make it per block _height_, not per tick: At every height, a block will eventually be proposed and finalized. It might just be in a different tick than expected. The closer the delivery height, the more likely it becomes that the block will be at a particular timestamp, which will be reflected in the price the futures are traded at.
> 

I think it would work either with ticks or block height, since in any final history there is at most one block at each position, it's just that the tick one is not guaranteed to exist even in a live chain. One could sell futures for either, in principle. But I do agree that whatever is sold must correspond to space in a particular block in the eventual finalized history.

I think this proposal, currently, is a good starting point, but needs significant additions to the narrative (the missing guide-level explanation) and probably a rework of the reference guide to deal less with how specific parameters work and more with "how is the auction implemented." For instance, it seems the escrow thing should be the central concept.

afck:
> it's just that the tick one is not guaranteed to exist even in a live chain

Exactly, but the height _is_. So if you buy futures for block 42, you know that eventually there will be a block 42; it can't be skipped.

osolmaz:
> I'd make it per block height, not per tick: At every height, a block will eventually be proposed and finalized. It might just be in a different tick than expected. The closer the delivery height, the more likely it becomes that the block will be at a particular timestamp, which will be reflected in the price the futures are traded at.

I see the point. I think there is a trade-off here. Maybe we could have both per-block-height gas futures, and per-period futures seamlessly working together? Futures could have different time-specificities, such as

- Per-day (low time-specificity)
- Per-hour (higher)
- Per-block-height (highest)

and block validity conditions would be derived from the outcome of the auction. Then, futures with high time-specificity would be prioritized when the validator is filling the block. They would also be more expensive when being auctioned off or traded.

Per-block-height futures would be used by dapp maintainers/users with *very large* volume, whose transactions get included in roughly every block. Futures with lower time specificity would then be favored by dapps/users with lower volume, but who still want to hedge themselves against high gas prices. (I insist on futures with lower time-specificity, because after using Ethereum and observing others use it, I know what the users need)

I think this is technically possible. We would probably need to specify per-hour and per-day futures also in terms of block-height, rather than ticks, to make them work together with per-block-height futures. What do you think?

afck:
Yes, that sounds good. :+1: 

> specify per-hour and per-day futures also in terms of block-height

I'm not even sure that's necessary.

osolmaz:
I applied the changes we discussed earlier.

text/0017-gas-futures.md
```diff
@@ -0,0 +1,144 @@
...
+
+which is placed in escrow, until the date of delivery. This mints a [non-fungible token](https://en.wikipedia.org/wiki/Non-fungible_token) (NFT) which we call the *futures escrow token* (FET). A FET represents a user's ownership of the BST, and stores the information
+
+```python
+(
+    owner, # The user who bought the futures
+    symbol, # BST symbol e.g. BST-2020-09-01-01000. Can also be any type of reference.
+    amount, # The amount of futures bought
+    price, # The auction price at the time of purchase
+    total_gas_used, # Amount of gas used with this FET
```
afck, line 95: I don't think we need that anymore, since the whole FET now refers to only a single block?

text/0017-gas-futures.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Auction
+
+The auction for `GAS-YYYY-MM-DD-NNNNN` begins at tick `X - GAS_AUCTION_PERIOD * TICKS_IN_DAY` where `X` is the tick corresponding to `YYYY-MM-DD` 00:00 in UTC, and `TICKS_IN_DAY = 24*60*60*1000` is the number of ticks in one day. The auction can go on until the whole quota is sold off.
+
+The auction price is determined by the validators collectively, who announce the price they are willing to accept. The futures price is computed by taking a weighted average:
+
+```python
+auction_price = sum(weight[v] * price[i] for v in validators) / sum(weight[v] for v in validators)
```
afck, line 70: As discussed, this should probably be a (weighted) median instead of average. But that's probably a question for a separate discussion.

afck (approved):
Looks good to me.

A possible mechanism against censorship:
Allow earlier blocks to refer to a transaction, i.e. if you have a future for block 1000, you can already get it "included" — but not executed — in block 998, to force the proposer of block 1000 to execute it: They can't pretend they haven't seen it.
That would potentially triple the _size_ of the futures portion of 998, since you'd have to download transactions for future blocks 999 and 1000 as dependencies as well, even though you're not executing them yet. However, the _CPU cost_ would only occur at block 1000. And censoring would require three validators to collaborate, instead of just one.

zie1ony:
My thoughts:
1. Per-block gas futures are useful only to those who want to do very precise actions like oracles. That doesn't help with scenarios where you know that you will make approximately 1000 transfers per day, but you don't know exactly when, because that depends on your customers' (random) actions. I think this is the problem we want to be solving. Another case is that we can't set a block target for a deploy.
2. The design would be cleaner and easier to use inside the smart contract if there is no FET concept. I would keep just the BST transferable tokens.
3. This plays well with [CEP-15#Delegated Payment](https://github.com/zie1ony/ceps/blob/payments/text/0015-gas-payments.md#delegeted-payment) idea. It would allow to offer cheaper (then the current price) transactions using unused gas or even allow others to build whole services that relay transactions for less.

xcthulhu:
Shelving this idea for now, since it does not make sense with the way transactions are prioritized.

PR #18, opened 2020-09-28
Token standard

zie1ony:
A standard interface for the CSPR and the custom made tokens.
[Rendered](https://github.com/zie1ony/ceps/blob/token/text/0018-token-standard.md)


EdHastingsCasperLabs (approved):
effectively accepted as there are implementations of it in use on-chain

PR #20, opened 2020-11-09
Block Proposer improvements CEP

fizyk20:
Rendered: https://github.com/fizyk20/ceps/blob/block-proposer-improvements/text/0020-block-proposer-improvements.md

### DIFF
--- a/text/0020-block-proposer-improvements.md
+++ b/text/0020-block-proposer-improvements.md
@@ -27,9 +27,23 @@ A separate solution for each of these two issues will be described below.
 
 ### Seeing finalization before proposal
 
-The solution to the first issue should be fairly straightforward: buffer finalized blocks that haven't been proposed as far as we know, and then handle both proposal and finalization when we see the proposal.
+#### Option 1
+
+The first potential solution to the first issue is fairly straightforward: buffer finalized blocks that haven't been proposed as far as we know, and then handle both proposal and finalization when we see the proposal.
 
 Since both the proposal and finalization events ultimately come from our own Consensus component, and we won't finalize a block we haven't seen proposed, we can expect that when the Block Proposer sees a finalization of a block that apparently wasn't proposed, it will also see the proposal shortly. This means we don't even have to take special care to prune old finalized blocks from the buffer, since we should never see a block that has been finalized, but is never proposed.
+
+However, this still doesn't solve the issue of orphaned blocks. Since we have no orphaned block detection currently, a block that is orphaned will stay in the `proposed` collection forever, thus making nodes never propose the deploys that were contained in them.
+
+Another pain point is that we would need a way to tell proto-blocks with identical deploy lists (like an empty list) from each other - otherwise, when we receive a notification about a block being finalized, we wouldn't know which proto-block it refers to, and we wouldn't be able to have multiple identical proto-blocks in the `proposed` collection. A way of doing that would be to pass in a parent block hash along with the proto-block.
+
+#### Option 2
+
+The second potential solution would be to not notify the Block Proposer about proposed blocks at all. The `past` parameter passed to it would contain the hashes of deploys contained in the blocks that haven't been finalized yet, instead of the hashes of blocks themselves. This would make the `past` list a few times longer, but would remove the need for the Block Proposer to know which blocks have been proposed - it could only care about finalized blocks in this approach.
+
+This option's another advantage would be removing the need for detection of orphaned blocks. If we don't record proposals, we won't have to "restore" the deploys from there when a block is orphaned.
+
+Yet another advantage is the lack of a need to distinguish between identical deploy lists. We wouldn't care if multiple proto-blocks with empty lists have been proposed, we would only care about the actual set of deploys that are in the past of the block to be proposed.
 
 ### Request for a block before seeing another validator's proposal
 
@@ -47,6 +61,8 @@ Even with this solution in place, we could actually still experience problems. C
 
 ### Seeing finalization before proposal
 
+#### Option 1
+
 When receiving `FinalizedProtoBlock(block)`:
 
 - check that `block.hash()` is in `self.state.proposed`
@@ -58,6 +74,14 @@ When receiving `ProposedProtoBlock(block)`:
 - check if `block.hash()` is in `self.state.finalization_queue`
 - if yes, remove it from `finalization_queue` and proceed as if we received both `ProposedProtoBlock` and `FinalizedProtoBlock`
 - if no, proceed in the current manner
+
+The last reuired part would be orphaned block detection, but it is not an improvement to the Block Proposer, but rather a missing feature in Consensus, and so is beyond the scope of this CEP.
+
+#### Option 2
+
+- Remove `self.state.proposed`, `added_block()`, `orphaned_block()`, `Event::ProposedProtoBlock` and consensus announcement about a proposed proto block
+- In `remaining_deploys`, change `past_blocks` to `past_deploys` which would be a set of deploy hashes instead of block hashes
+- Change the calculation of `past_deploys` in the method to append the deploys from finalized blocks to `past_deploys`
 
 ### Request for a block before seeing another validator's proposal
 
--- a/text/0020-block-proposer-improvements.md
+++ b/text/0020-block-proposer-improvements.md
@@ -27,23 +27,11 @@ A separate solution for each of these two issues will be described below.
 
 ### Seeing finalization before proposal
 
-#### Option 1
+The suggested solution would be to not notify the Block Proposer about proposed blocks at all. The `past` parameter passed to it would contain the hashes of deploys contained in the blocks that haven't been finalized yet, instead of the hashes of blocks themselves. This would make the `past` list a few times longer, but would remove the need for the Block Proposer to know which blocks have been proposed - it could only care about finalized blocks in this approach.
 
-The first potential solution to the first issue is fairly straightforward: buffer finalized blocks that haven't been proposed as far as we know, and then handle both proposal and finalization when we see the proposal.
+One of this solution's advantages would be removing the need for detection of orphaned blocks. If we don't record proposals, we won't have to "restore" the deploys from there when a block is orphaned.
 
-Since both the proposal and finalization events ultimately come from our own Consensus component, and we won't finalize a block we haven't seen proposed, we can expect that when the Block Proposer sees a finalization of a block that apparently wasn't proposed, it will also see the proposal shortly. This means we don't even have to take special care to prune old finalized blocks from the buffer, since we should never see a block that has been finalized, but is never proposed.
-
-However, this still doesn't solve the issue of orphaned blocks. Since we have no orphaned block detection currently, a block that is orphaned will stay in the `proposed` collection forever, thus making nodes never propose the deploys that were contained in them.
-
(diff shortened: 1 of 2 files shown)


text/0020-block-proposer-improvements.md
```diff
@@ -0,0 +1,108 @@
...
+
+A separate solution for each of these two issues will be described below.
+
+## Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+### Seeing finalization before proposal
+
+The solution to the first issue should be fairly straightforward: buffer finalized blocks that haven't been proposed as far as we know, and then handle both proposal and finalization when we see the proposal.
```
afck, line 30: What if we have e.g. lots of empty blocks? They will all have identical `ProtoBlock`s, and they might all get finalized. Will that cause us to remove the "empty" entry from the proposed blocks, and then buffer it forever if another empty block gets finalized?

(Maybe we should rename `ProtoBlock` to `DeployList` as part of this change?)

text/0020-block-proposer-improvements.md
```diff
@@ -0,0 +1,108 @@
...
+
+When receiving a `ListForInclusion` request:
+
+- check if `last_finalized_block` is in `self.state.finalized`
+- if yes, proceed in the current manner
+- if not, buffer the request in a new field `self.state.request_queue`
+
+When receiving a `FinalizedProtoBlock` event:
+
+- check if the block's parent is in `self.state.finalized` (in addition to other checks that are in place)
```
afck, line 74: That means we'll have to store `BlockHash` instead of `ProtoBlock` in that case, does it? (`ProtoBlock` doesn't have an explicit parent field.)
Which is fine, I think, because `ProtoBlock` can be computed from it.

text/0020-block-proposer-improvements.md
```diff
@@ -0,0 +1,108 @@
...
+- if yes, proceed; re-check whether blocks in `self.state.finalization_queue` can be finalized and finalize those that can
+- if no, buffer the block in `self.state.finalization_queue`
+
```
afck, line 77: We also have to finally implement detecting orphaned blocks (otherwise they stay in `proposed` forever), in addition to passing the ancestors up to and including the last finalized block to the block proposer, from the consensus side.

text/0020-block-proposer-improvements.md
```diff
@@ -0,0 +1,108 @@
...
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+This adds more state to the `BlockProposer`, which might be undesirable. However, the issues caused by out-of-order events are pressing and the benefits likely outweigh the cost.
+
+## Rationale and alternatives
```
afck, line 84: To my mind, the main alternative is to get rid of the `proposed` collection. I don't understand why we need it anymore.
Because whenever we propose deploys, we now know all ancestors, so we can just exclude those. There's no harm if they are still in the `pending` collection. That would also resolve the issue with finalizing a block that wasn't proposed: The block proposer wouldn't have to care at all about the event that a new block has been proposed. And we wouldn't have to detect orphaned blocks.

So basically we'd just have the `pending` and `finalized` collections (even the latter could just contain deploys, not blocks).
* Whenever we're asked to propose deploys, we take stuff from `pending`, and exclude all ancestors.
* If we receive a new incoming deploy, we add it to `pending` except if it's already in `finalized`.
* If we're told about a newly finalized block we add its deploys to `finalized` and remove them from `pending`. (No error if they aren't there!)
* We also track the last finalized block, as you propose above (and check for its parent etc.).

text/0020-block-proposer-improvements.md
fizyk20: We should know the parent block at the moment of proposal, shouldn't we? I mean, it should at least be possible to get the parent block hash from somewhere. We could then make the distinction between different blocks with the same `ProtoBlock`s by parent hash.

Good point, though.

And I have nothing against the renaming :+1:

text/0020-block-proposer-improvements.md
goral09: I think one of the main reasons to have `proposed` collection was so that we don't send all deploy hashes from the consensus to the block executor, but refer to them by the proto hash which is shorter and should uniquely identify the deploys.

afck: But then we _are_ sending it all deploy hashes, aren't we?
The difference is that currently we send them when we learn about a _proposed_ block, and with my proposal, we would only send them when _finalizing_ a block.

text/0020-block-proposer-improvements.md
afck: It would certainly be possible to put a parent hash into the `DeployList`; however, it would be kind of redundant and I think it would be far from trivial, given how the code is currently organized.

We currently don't have the concept of a proto block/deploy list parent. Introducing it would mean to keep track of almost a "parallel" preliminary chain of deploy lists. Since we currently need to be able to recreate a deploy list from a linear chain block, we'd end up putting that second "deploy list parent hash" into linear chain blocks, in addition to the proper parent _block_ hash…

Anyway, I don't think we should do that.

fizyk20: I don't mean putting the parent hash into the `DeployList`. Rather, when we notify the Block Proposer about a proposal or finalization, we supply the parent block hash _along_ with the deploy list. Just so that we know that _this_ empty deploy list is a _different_ deploy list from _that_ empty deploy list. Or something along these lines.

text/0020-block-proposer-improvements.md
fizyk20: The idea here was to compress the data slightly, I think. But the idea was born when we were still considering sending the _complete_ past with every proposal request, and in such a case the list of block hashes should be much more efficient than the list of deploy hashes. However, if we are sending just the not-yet-finalized part of the past, we can just as well make it a list of the deploy hashes, I think. So then we will know that we should include all the deploys that aren't in finalized blocks or in this list - and we don't need to know about which blocks were proposed :thinking:

goral09: I think the idea has always been to send only the _not-yet-finalized_ part and rely on the fact that block proposer knows the finalized part.

text/0020-block-proposer-improvements.md
afck: But if it's three subsequent empty blocks, we have two identical pairs of deploy lists.
I guess we could also use the block height or something. Nor sure.

text/0020-block-proposer-improvements.md
afck: Ah, right, good point! A list of block hashes would be smaller, of course.
But it would also require unique identifiers for deploy lists in different blocks, since we don't have block hashes yet at this point.

text/0020-block-proposer-improvements.md
xcthulhu: I am not sure about the architecture, so please forgive me if the following question is naive.

Can I spam finalized blocks (perhaps dated in the future)?  What would stop me from flooding the proposed `finalization_queue`?

xcthulhu: > I don't mean putting the parent hash into the `DeployList`. Rather, when we notify the Block Proposer about a proposal or finalization, we supply the parent block hash along with the deploy list. Just so that we know that this empty deploy list is a different deploy list from that empty deploy list.

Also: if I notified the Block Proposer with `(parent_block_hash, deploylist1)` and `(parent_block_hash, deploylist2)`, would it break anything?

(thank you again for your patience with my dumb questions)

xcthulhu: Chatting with @afck , we can't spam finalized blocks because they have 2/3rds of the validator signatures

text/0020-block-proposer-improvements.md
afck: I think with what I'm suggesting above we'd also reduce the surface between the consensus component and the rest: We could get rid of [`announce_proposed_proto_block`](https://github.com/CasperLabs/casper-node/blob/3516f68a5f86db080fb4495a39f58929b5d81f25/node/src/effect.rs#L856). (Which is currently [called too early anyway](https://github.com/CasperLabs/casper-node/blob/3516f68a5f86db080fb4495a39f58929b5d81f25/node/src/components/consensus/era_supervisor.rs#L546), I think: It needs to wait for the evidence for accusations… I made a note of it.)

afck: (I created https://github.com/CasperLabs/casper-node/pull/525 to fix that. :point_up_2:)

text/0020-block-proposer-improvements.md
```diff
@@ -59,6 +75,14 @@ When receiving `ProposedProtoBlock(block)`:
 - if yes, remove it from `finalization_queue` and proceed as if we received both `ProposedProtoBlock` and `FinalizedProtoBlock`
 - if no, proceed in the current manner
 
+The last reuired part would be orphaned block detection, but it is not an improvement to the Block Proposer, but rather a missing feature in Consensus, and so is beyond the scope of this CEP.
```
afck, line 78:
```suggestion
The last required part would be orphaned block detection, but it is not an improvement to the Block Proposer, but rather a missing feature in Consensus, and so is beyond the scope of this CEP.
```

afck (approved):
I'm in favor of option 2.

Or maybe option 1½:
The compression aspect could still be preserved, if we keep the "proposed block" notification but use it only for that purpose: Introducing a new abbreviation for a list of deploys. The block proposer wouldn't move any deploys around at this point: It would just take note of this new list and its hash. That way, duplicated lists (e.g. multiple empty blocks) don't cause issues.
The only downside would be that we'd still need a queue: If we receive any request or finalization notice about a proto block hash we don't know yet, we have to wait for the corresponding proposal.

On the other hand, maybe that's an optimization we could do much later, and just going with option 2 for now would be simpler.
(The optimization wouldn't even change the protocol; no fork or anything.)

Anyway, the CEP should probably contain a decision instead of two options before being merged, so let's wait for others to chime in before merging.

goral09:
I agree with Andreas. Depending on the scope creep we can either implement the optimization right away or not.

goral09 (approved):

text/0020-block-proposer-improvements.md
```diff
@@ -0,0 +1,110 @@
...
+
+The solution for this issue could be to include the hash of the last finalized block in the request for deploys. The Block Proposer could then check if it already has that block as finalized, and if not, it would buffer the request and complete it when it receives the notification of the relevant block's finalization.
+
+Even with this solution in place, we could actually still experience problems. Consensus could finalize multiple blocks at once, and if the events aren't guaranteed to arrive in order, the Block Proposer could be notified of the finalization of the last block before some others. In such a case, it would still have an incomplete picture of the new block's past. To prevent this, we would have to only handle notifications about finalized blocks if the block's parent has already been finalized (and buffer finalization notifications that don't satisfy this condition, until it is satisfied).
```
goral09, line 44: This looks to me like an unresolved question/problem. Is the implementation of this CEP addressing the issue?

goral09: It seems like it is addressed under [**Request for a block before seeing another validator's proposal**](https://github.com/CasperLabs/ceps/pull/20/files#diff-593efcb8ed59381bcf184107760c740e5afae65aa37c9d08c487f5804fd29de4R56). It's just that the language in this paragraph makes it sounds as if it's still an open question.

PR #21, opened 2020-11-11
State pruning and fast syncing

marc-casperlabs:
[Rendered](https://github.com/marc-casperlabs/ceps/blob/state-pruning-and-fast-syncing/text/0021-state-pruning-and-fast-syncing.md)


xcthulhu:
I'm afraid there's some very high priority work for the security audit that's happening right now.  This report is very long, and I am afraid I don't have time to properly review it this week.

Fraser999 (commented):
First pass - just nitpicks/typos.

text/0021-state-pruning-and-fast-syncing.md
```diff
@@ -0,0 +1,240 @@
...
+
+The "fast" part of fast-syncing, as compared to the status quo, is realized from two factors, namely
+
+* **side-loading** data from outside the network through external sources and
+* **transfering** state, namely trie-store state, which is currently not transferrable at all.
+
+### Side-loading
+
+We create a common trait for Blocks, Deploys or any other transferable object that states how it can be serialized and defines a hash function for its serialized representation, as well as a key to separate the hash space. As an example, a block whose hash is `5A4A55A8CA87B584960DCB0673DC1260292991EC6D2B8F214701B5224FB141D2` would be adressable as `blockv1_5A4A55A8CA87B584960DCB0673DC1260292991EC6D2B8F214701B5224FB141D2`. The `blockv1` denotes a complete scheme, that is it also pins the serialization format. While it is not necessary to separate the hash value space to avoid collisions, the prefix allows for multiple representations, should the format of a serialization or an internal data structure change at some point. We refer to data structures that satisfy this condition as *objects* from here on out.
```
Fraser999, line 92: I don't see a need for the prefix.  Are you trying to support retrospectively changing serialization formats for historic blocks?

Since we have a protocol version, I'd expect we'd be able to define all necessary systemic changes via that one versioning mechanism.  So, if a block was stored under protocol version 1.0.0, we definitively know the composition/structure of the block and its serialization format.

Fraser999, line 92: I think we should cleanly separate retrieval of a block, deploy, trie node from a peer vs retrieval from outside the network.  There is no need to have p2p message sizes increased by forcing URL-friendly representations of hashes.

text/0021-state-pruning-and-fast-syncing.md
```diff
@@ -0,0 +1,240 @@
...
+
+**Note**: The defunct CEP3 proposed hashing the serialized representation of an object instead of the object itself, which makes it a lot easier for external stores to handle these, as integrity can be checked without deserializing first. However, hashing the internal object minimizes the impact on the existing codebase instead.
+
+We can then define a generic mechanism for requesting these objects from other nodes. The true value however is in being able to serialize these to static external storage.
```
Fraser999, line 96: We already have a generic mechanism for requesting these objects from other nodes.  I'm not a fan of it, but it's there.

text/0021-state-pruning-and-fast-syncing.md
```diff
@@ -0,0 +1,240 @@
...
+
+### S3, HTTP and read/write, read-only stores
+
+As objects are unchanging, they can be readily be side-loaded from an HTTP-server that acts as a read-only store. As an example, should a node need to obtain the block `5A4A55A8CA87B584960DCB0673DC1260292991EC6D2B8F214701B5224FB141D2`, it can follow the following steps:
+
+1. Check one or more configured external HTTP stores by calling `GET store/objects/5A4A55A8CA87B584960DCB0673DC1260292991EC6D2B8F214701B5224FB141D2`
```
Fraser999, line 102: I'm not sure whether we should be favouring our own network as the first call.  I guess it depends upon whether we see our network as the primary storage source for current data with outside storage being a backup, or vice-versa.  I've been used to the former view, but maybe I need to rethink.

Long-term, I'm not clear on whether we see this off-network storage as a permanent solution or if it's a stop-gap until we build an archival system which would make our network self-reliant and crucially, not dependent on centralised solutions.

text/0021-state-pruning-and-fast-syncing.md
```diff
@@ -0,0 +1,240 @@
...
+1. If the request is a hit, try to download the object, deserialize it and validate the hash. Should any of these operations fail, discard and continue. Otherwise return.
+1. Ask the network (typically a node is known that is required to have it) for the object in the same manner.
+
+In the same manner, an S3-compatible API can be employed. Since many S3 providers allow making stored objects public, it would be highly recommendable to use a similar structure for the keys, as a result, every object uploaded to S3 automatically becomes read-only fetchable by other nodes that have access to the S3 bucket in real-time.
```
Fraser999, line 106: > it would be highly recommendable to use a similar structure for the keys
There's no need for the p2p messages to have URL-friendly keys.  It's sufficient to be able to map between these easily.

text/0021-state-pruning-and-fast-syncing.md
```diff
@@ -0,0 +1,240 @@
...
+
+Nodes configured to keep the external store up-to-date will, upon receiving a new object, attempt to upload it to the store, by first checking if it exists already and uploading it if it does not.
```
Fraser999, line 108: This needs a little more detail I believe.  It's probably fine for deploys, which only "change" once in their lifetimes (when they are executed - and even this might be devolved to the event store so that deploys are permanently immutable from the moment they're provided to the network by a client).  However, for blocks, signatures will be appended as and when they're received from validators.  We'd presumably want to avoid many duplicate updates to a block.

text/0021-state-pruning-and-fast-syncing.md
```diff
@@ -0,0 +1,240 @@
...
+
+Objects like blocks, deploys and similar are trivially convertable into objects. For the trie store it should also be possible, as every vertex on the tree is either a leaf, a node or an extension. The hash matches what the store calls a pointer, thus a recovery of a root hash `4813494D137E1631BBA301D5ACAB6E7BB7AA74CE1185D456565EF51D737677B2` from an external store or the network would look like this:
+
+0. Initialize the retrieval queue with `4813494D137E1631BBA301D5ACAB6E7BB7AA74CE1185D456565EF51D737677B2`.
+0. For every object in the retrieval queue, remove those already in the trie store. If the queue is now empty, exit.
+0. Request all objects though the trie scheme, e.g. `triev1_4813494D137E1631BBA301D5ACAB6E7BB7AA74CE1185D456565EF51D737677B2`, which should deserialize to a `Trie::Node`.
+0. After successful deserialization, add all descendants of retrieved object into the retrieval queue, or abort with an error.
+0. Go to 1.
+
+### Optimization of retrieval
```
Fraser999, line 122: I believe this section belongs in the "Future possibilities" section.

text/0021-state-pruning-and-fast-syncing.md
```diff
@@ -0,0 +1,240 @@
...
+
+Retrieval of any object can be optimized by batching, which will not be discussed in depth here. The general idea is to request multiple objects in a single request, and/or provide pre-assembled pack files that group data that is often requested at the same time, like a block and its deploys or the children of a subtree of a global state hash.
+
+This is an optimization that should be deferred for now though.
+
+### Putting everything together
+
+With a combination of the proposals above, a node's joining process from a trusted hash ideally becomes this:
+
+1. Download the block given in the trusted hash, sped-up through external caches.
```
Fraser999, line 132: Is it necessarily faster?  Even if we're connected to a peer on a local network say?

Fraser999 (commented):
This is much more acceptable to me than its predecessor, primarily as it appears to have a much lower impact on the existing code.

However, I still think that the efforts to homogenise all the data types mentioned here are pointless and will actually be detrimental to clean, clear code.  In other words, I'd prefer to not start with a `Scheme` trait and likewise it seems pointless to format requests to get/put data inside the network in the same way as those outside.

I'd think we'd benefit from a more measured, incremental approach - e.g.
1. implement pruning of blocks and deploys
2. implement pruning of global state
3. implement offloading and retrieval to/from outside storage provider(s)
4. implement fast sync

Some of this work could be parallelised depending upon resources, but in any case, it should become apparent during this work whether the differences in handling the various data types make pushing for a unifying trait worthwhile or not.  My expectation is that it won't.

text/0021-state-pruning-and-fast-syncing.md
```diff
@@ -0,0 +1,240 @@
...
+
+This leaves us with the problem of deciding which root hashes to keep, which will be addressed later.
+
+### Storage pruning
+
+Pruning of the storage offers more substantial savings than the trie store, as most disk space is to consumed by deploys. Each block has associated data (deploys, finality signatures, etc.) that is unlikely to be shared by other blocks, as deploys are only added to one block and finality signatures are also unique. Evicting blocks from the store should thus free up space proportionally on average.
+
+By deciding to "cut" the block chain at a specific parent, we can apply a similar garbage collection algorithm:
+
+1. Decide which blocks to keep and add them to the "objects to keep" set (*i.e.* a starting block and all descendents).
```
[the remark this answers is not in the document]
marc-casperlabs: I don't want to rely on indices for general garbage collection to work. The "objects to keep" set also has an advantage, it makes it possible to garbage collect a running system: When the GC is running, every newly stored object is added to the "objects to keep" set when inserted and thus not touched.

Or put differently, like specifying what to keep and removing the rest instead of the opposite, since it clears out any strays :)

EdHastingsCasperLabs (approved):
So...I agree with the broad strokes of this and rather than get run aground quibbling over specific details I would like to see an MVP defined that achieves the fast syncing and general storage pruning capabilities, leaving global state pruning and other goals as a TBD for future exploration.

goral09:
I wanted to chime in on the "fast syncing" part. The next "fast syncing" milestone shouldn't rely on speeding up the download process by parallelizing the download over multiple high-throughput low-latency data sources. Syncing will be fast b/c instead of downloading the whole chain nodes will download only specific, singular blocks of the linear chain until they find the latest state they can trust. Then, they will download the Global State at that block. The step in this process that needs high-throughput and low-latency is downloading of a _single block_ and downloading of a GlobalState and _specific state hash_. There shouldn't be any need to download _everything_.

text/0021-state-pruning-and-fast-syncing.md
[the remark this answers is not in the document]
henrytill: What is preventing us from running GC in the background @Fraser999's solution?  

One drawback of the keep list based solution is that it would need to be persisted to disk in the event that the node is stopped and restarted.

text/0021-state-pruning-and-fast-syncing.md
```diff
@@ -0,0 +1,240 @@
...
+1. For each block, visit all of its dependencies (deploys, finality signatures) and add them to an "objects to keep" set.
+1. Visit each block, deploy and finality signature, remove them if not inside the "objects to keep" set.
+
+In a manner similar to the trie store, we can employ a mark-and-sweep version of this pruning algorithm as well.
+
+This again reduces the problem on selecting which blocks to keep.
```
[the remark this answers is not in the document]
henrytill: Agree with @Fraser999, not seeing the problem here.

text/0021-state-pruning-and-fast-syncing.md
henrytill: I'm not seeing the need either.  

I would assume every data fetch would starts with the getting a block, and the block itself has protocol version field.  As long as we guarantee that the protocol version will always have a predictable/fixed location and format in the binary format a block, all subsequent fetches,  be they deploys or trie nodes, will be able to use the block's protocol version to determine which deserializer to use.

text/0021-state-pruning-and-fast-syncing.md
henrytill: Agreed, we should be looking at how to iterate/improve on that rather than inventing a new solution.

text/0021-state-pruning-and-fast-syncing.md
henrytill: > I'm not sure whether we should be favouring our own network as the first call. I guess it depends upon whether we see our network as the primary storage source for current data with outside storage being a backup, or vice-versa. I've been used to the former view, but maybe I need to rethink.

I've been used to the former view as well.

henrytill (commented):
Quoting @Fraser999, as he's captured a lot of my own thoughts in his review:

> This is much more acceptable to me than its predecessor, primarily as it appears to have a much lower impact on the existing code.

Very much agreed here.

> However, I still think that the efforts to homogenise all the data types mentioned here are pointless and will actually be detrimental to clean, clear code. In other words, I'd prefer to not start with a `Scheme` trait and likewise it seems pointless to format requests to get/put data inside the network in the same way as those outside.

Also agree here.  See my earlier [comment](https://github.com/CasperLabs/ceps/pull/21#discussion_r536342030).

> I'd think we'd benefit from a more measured, incremental approach - e.g.
> 
>     1. implement pruning of blocks and deploys
> 
>     2. implement pruning of global state
> 
>     3. implement offloading and retrieval to/from outside storage provider(s)
> 
>     4. implement fast sync
> 
> 
> Some of this work could be parallelised depending upon resources, but in any case, it should become apparent during this work whether the differences in handling the various data types make pushing for a unifying trait worthwhile or not. My expectation is that it won't.

Very much agreed here as well.

I will also register my vote in support of @Fraser999's proposed iterative/indexed-based approach to pruning.

text/0021-state-pruning-and-fast-syncing.md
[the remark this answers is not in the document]
marc-casperlabs: > What is the problem? Isn't it a very simple matter of all blocks with a height less than H?

It's mainly to separate garbage collection from a specific implementation. The trie store, to the best of my knowledge, does not have blocks, but substituting root hashes here, we can think of schemes in the future where we do not want to prune everything up to a certain height. It is conceivable to keep some historic states around, for example, but not all of them (e.g. one per week), to be able to reconstruct intermediate states quickly.

text/0021-state-pruning-and-fast-syncing.md
marc-casperlabs: > I don't see a need for the prefix. Are you trying to support retrospectively changing serialization formats for historic blocks?

Data at rest cannot be assumed to be updated, I am thinking of external stores as basically write-only (e.g. a huge, historical S3 bucket). It is conceivable that data might be added twice in different formats, should a need arise to change it.

If we want to version instead, there is an alternative scheme: Store every object at an endpoint like

```
GET /objects/v1.0.0/5A4A55A8CA87B584960DCB0673DC1260292991EC6D2B8F214701B5224FB141D2
```

I am assuming that a change of serialization format automatically increments a major version, so we might get away with

```
GET /objects/v1/5A4A55A8CA87B584960DCB0673DC1260292991EC6D2B8F214701B5224FB141D2
```

Now if the format changes, we would have to create a new object at 

```
GET /objects/v2/5A4A55A8CA87B584960DCB0673DC1260292991EC6D2B8F214701B5224FB141D2
```

via a conversion script. For objects whose serialization format hasn't changed, this would hopefully just be another URL where it is exposed on, without changing the underlying data.

> I would assume every data fetch would starts with the getting a block, and the block itself has protocol version field. As long as we guarantee that the protocol version will always have a predictable/fixed location and format in the binary format a block, all subsequent fetches, be they deploys or trie nodes, will be able to use the block's protocol version to determine which deserializer to use.

I believe that's just inlining the otherwise external version tag. It relies on the fact that a version tag is found at the same place in every object, which I find a bit brittle - should we ever decide to change things in the future, this might cause problems.

There is also the case where a serialization format for an object changes, without its hash (based on values) changing, so these objects might be transferred across the wire differently even though they are the same.

marc-casperlabs: > I think we should cleanly separate retrieval of a block, deploy, trie node from a peer vs retrieval from outside the network. 

I'm not sure we're not making our lives extra difficult here. By nature of coming from different components and different transports, objects are already separately retrieved, what extra separation do we need?

> There is no need to have p2p message sizes increased by forcing URL-friendly representations of hashes.

I may have phrased this poorly, but there address is just for URL purposes on outside stores - there's no need to send the hash as a hexstring in our own wireformat.

text/0021-state-pruning-and-fast-syncing.md
marc-casperlabs: > We already have a generic mechanism for requesting these objects from other nodes. I'm not a fan of it, but it's there.

The current implementation is, politely put, bad. I would remove it rather sooner than later, as it completely violates the whole reactor design and has been an issue requiring workarounds in both reactor refactoring as well as the storage component rewrite. It should have never been added in the first place and yes, I take full responsibility for that since I was the one that approved the PR that added it... Let me atone for my past sins, please! :)

text/0021-state-pruning-and-fast-syncing.md
marc-casperlabs: I think this is sort of the wrong perspective - we should be looking at these as authoritative vs optional, instead of primary vs secondary. Luckily we don't have to deal with conflicting values, since everything is hashed, so the "authority" part is just regarding completeness.

My view is as follows: The decentralized network is the ultimate authority on everything, at it should always (somehow) contain everything considered keepable, while other sources are caches. Put differently, a primary source would always be asked first and considered an authority, while a secondary wouldn't. A cache splits these concerns, one will always want to ask a cache first (speedier) and fall back on the actual authoritative source on a miss.

So in short: Ask the off-network storage first, but always ensure that the network can act as a fallback. In the case of a failure of all external off-network storage, things would work a lot slower, but still work (and external stores could be rebuilt from it).

text/0021-state-pruning-and-fast-syncing.md
marc-casperlabs: "Structure" is meant to at least mean "tag + hash" here, not necessarily the hex-encoding. It could be encoded binary as well, I am not entirely sure of the savings though, especially if stream compression is ever to be added.

text/0021-state-pruning-and-fast-syncing.md
marc-casperlabs: CEP21 assumes that there are only immutable objects in the store. I am 100% against storing block signatures long-term as mutations of the blocks, but instead coming up with a scheme that stores them immutable next to them. We should not be storing anything mutable at all.

text/0021-state-pruning-and-fast-syncing.md
marc-casperlabs: I'd say yes. The peer-on-local network is likely the exception for most operators. If so, this could be special cased or configurable.

Anyone else will find it hard to compete with a commercial CDN (which depending on your ISP might also be on the local network, although not financially :))

text/0021-state-pruning-and-fast-syncing.md
marc-casperlabs: Fair enough, it has been moved!

marc-casperlabs:
> The next "fast syncing" milestone shouldn't rely on speeding up the download process by parallelizing the download over multiple high-throughput low-latency data sources.

That is true for the _next_ milestone, but ultimately I want both!

Regarding @henrytill and @Fraser999 's comments, I think this CEP should be split into multiple ones:

1. Restructuring of objects in storage/outside storage/on the wire. Making everything immutable and cleaning up the networking part (`GetRequest` must go) I see as a prerequisite for further work.
2. Based on that, fast syncing (which mainly deals making global state transferable).
3. Pruning storage and trie store.
4. Cherry on top, external caches.
5. Potentially integrating consensus into this new framework

Of these, 1. Is a blocker, but I believe 2-4 can be parallelized, while 5 needs a lot of extra design work.

xcthulhu:

PR #22, opened 2020-11-23
Gas spot market

AlexanderLimonov:
Initial outline, needs filling in for exact processes + clarification of fuzzy heuristics expressed on Slack regarding balance checks

[Rendered](https://github.com/CasperLabs/ceps/blob/Gas_spot_market/text/0022-gas-spot-market.md)

### DIFF
--- a/text/0000-gas-spot-market.md
+++ b/text/0000-gas-spot-market.md
@@ -6,31 +6,102 @@
 
 CEP PR: [casperlabs/ceps#0001](https://github.com/casperlabs/ceps/pull/0001)
 
+The present CEP is intended to guide the development of the gas allocation mechanism that will be present at mainnet launch, 
+later to be supplemented by a gas futures market and a fiat link. The CEP explains the concepts of willingness to pay (WTP), spot gas reservation,
+gas allocation, gas requirements estimation, refunds and price determination. Further, it addresses the separation of block gas allocation for regular deploys and WASM-less transfers, 
+balance checks and fallback purses. The present CEP is not intended to be a technical specification, but rather an explanation of the business logic
+and the economic imperatives leading to the proposed deploy lifecycle.
+
 ## Motivation
 
 [motivation]: #motivation
 
-(initial implementation of a transparent market for gas)
+The Delta testnet that is active at the time this CEP is being drafted does not have any meaningful gas pricing, although it does charge
+a nominal, hardcoded amount of motes per unit of gas. The reason we need to enable market pricing for gas is simply that gas represents
+time spent on computing the changes to global state, thus making it an approximate measure of a finite resource. This resource is finite
+because meaningful use of a blockchain requires imposition of what is effectively a soft time limit on block execution. Finite resources
+must be priced both because it prevents spam attacks on the blockchain and because it is a means to discover the highest return uses
+for the resource, which ultimately should be allocated to those willing to pay the most.
 
 ## Guide-level explanation
 
 [guide-level-explanation]: #guide-level-explanation
 
-### What do we mean by "gas price?"
+In this section, we will outline the basic concepts and UX for blockchain users.
+
+### Basic concepts
+
+#### Willingness to pay (WTP)
+
+Willingness to pay can be seen as analogous to "price" as informally used when discussing user bidding for gas. We will shortly
+address why "price" should not be used for the deploy parameter we are about to define. Willingness to pay is almost self-explanatory, 
+being the amount a user is willing to pay for a unit of gas. We are using the term pragmatically, not intending to mean that this is the
+*maximum* a user is willing to pay. In fact, under our scheme, bidding is almost certainly going to be strategic, with users "shading" their
+true willingness to pay, since the proposed allocation rule is analogous to a first-price auction.
+
+#### Spot gas reservation
+
+Spot gas reservation, conceptually, is the claim a user will lay on gas in a proposed block. Similarly to willingness to pay, it will be
+a deploy parameter. The user effectively buys this amount, priced at the stated willingness to pay, when included in a block. To soften
+the impact of poor gas requirements estimation, we will be implementing partial refunds for unused gas.
+
+#### Gas allocation
+
+Gas allocation with a greedy revenue-optimizing ordering of deploys is the core activity that this CEP proposes. It is envisioned that this
+rule will be hardcoded into the block proposer component. The rule will be that highest WTP deploys with the highest spot gas reservations
+be included first, subject to certain balance requirements. The charges arising from inclusion would be WTP multiplied by the spot gas reservation,
+less any refunds arising from incomplete use of reserved gas.
+
+#### What do we mean by "price?"
+
+We have previously said that "price" should not be used when referring to the parameter we will be calling willingness to pay. The reason
+for this is clarity of communication to external parties. Ultimately, "price" is a property of the market, not of individual discrete 
+claims on a resource comprising the demand for the resource. "How much does it cost per unit of gas to include a reasonably sized deploy?" is a question
+logically met with the answer "an epsilon above what the last deploy to make it into the last observed finalized block paid." However, because of uncertainty
+at this margin, stating a higher willingness to pay leads to a higher probability of inclusion, a behavior users would expect from blockchains like Ethereum.
+In fact, stating a willingness to pay above the observed "market price" directly amounts to paying extra for higher probability of inclusion, because of our allocation
+rule.
 
 ### What does the user see?
 
+The UI change for the user is the introduction (possibly by renaming) of a "willingness to pay" parameter and a "spot gas reservation" parameter. The UX
+becomes somewhat more complex, since it is up to the user to determine the gas requirements. Because of our consensus-before-execution model, we are unable
+to allocate gas on the fly. This means that to avoid underutilization of block gas, we must charge for the entire reserved amount (although with partial refunds, in recognition
+of difficulties with estimating gas use of complex deploys). It is expected that the users, and particularly high WTP users, will make their best effort to estimate gas
+requirements offline, or on our testnet, before submitting the deploys to the live network. We expect that, eventually, the community will develop tooling to make this estimation easier.
+Some of the potential solutions may large dApps contributing developer time and nodes to maintaining a robust testnet that shadows complex contracts deployed on the live chain, adaptation of
+our testing tools and application of empirical methods to observed gas use patterns.
+
 ### What does the protoblock proposer see?
 
-### How is the pricing resolved?
+The protoblock proposer, here a personification of the block proposer component, sees a deploy buffer, as usual. However, the block proposer will now be hardcoded to follow
+the simple greedy rule we outlined above, alongside with balance checks to prevent attacks by "forgetting one's wallet at home" (we will refer to this, more formally, as the "nothing-in-purse problem").
 
 ## Reference-level explanation
 
 [reference-level-explanation]: #reference-level-explanation
 
+### Chainspec parameters
+
+#### Block gas limit
+
+#### Share of gas dedicated to WASM-less transfers
+
 ### Deploy channels
 
 ### Lifecycle of a deploy
+
+#### Ordering
+
+#### Balance check
+
+#### Allocation
+
+#### Payment code execution
+
+#### Session code execution
+
+#### Refunds
 
 ## Drawbacks
 
--- a/text/0000-gas-spot-market.md
+++ b/text/0000-gas-spot-market.md
@@ -1,0 +1,64 @@
+# Spot market for gas
+
+## Summary
+
+[summary]: #summary
+
+CEP PR: [casperlabs/ceps#0001](https://github.com/casperlabs/ceps/pull/0001)
+
+## Motivation
+
+[motivation]: #motivation
+
+(initial implementation of a transparent market for gas)
+
+## Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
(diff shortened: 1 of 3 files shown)


text/0000-gas-spot-market.md
```diff
@@ -0,0 +1,135 @@
...
+rule.
+
+### What does the user see?
+
+The UI change for the user is the introduction (possibly by renaming) of a "willingness to pay" parameter and a "spot gas reservation" parameter. The UX
+becomes somewhat more complex, since it is up to the user to determine the gas requirements. Because of our consensus-before-execution model, we are unable
+to allocate gas on the fly. This means that to avoid underutilization of block gas, we must charge for the entire reserved amount (although with partial refunds, in recognition
+of difficulties with estimating gas use of complex deploys). It is expected that the users, and particularly high WTP users, will make their best effort to estimate gas
+requirements offline, or on our testnet, before submitting the deploys to the live network. We expect that, eventually, the community will develop tooling to make this estimation easier.
+Some of the potential solutions may large dApps contributing developer time and nodes to maintaining a robust testnet that shadows complex contracts deployed on the live chain, adaptation of
```
afck, line 72:
```suggestion
Some of the potential solutions may be large dApps contributing developer time and nodes to maintain a robust testnet that shadows complex contracts deployed on the live chain, adaptation of
```

afck (commented):
Looks good to me so far! :+1:

afck (commented):
Still looking good. :slightly_smiling_face:

afck (approved):

xcthulhu:
From my understanding, here is what is being proposed:

0. Both wasmless transfers and deploys need a gas price associated with them which will be paid, and validators will prioritize them based on this.
1. wasmless transfers will cost a fixed gas amount
2. Some amount of gas per block needs to be dedicated to wasmless transfers

@AlexanderLimonov is that right?

xcthulhu (approved):

AlexanderLimonov:
> 
> 
> From my understanding, here is what is being proposed:
> 
>     1. Both wasmless transfers and deploys need a gas price associated with them which will be paid, and validators will prioritize them based on this.
> 
>     2. wasmless transfers will cost a fixed gas amount
> 
>     3. Some amount of gas per block needs to be dedicated to wasmless transfers
> 
> 
> @AlexanderLimonov is that right?

Yes. On 2, my current understanding is that it's a host-side function we can time, giving us the cost.

henrytill:
Tagging @Fraser999 as it appears that there are things here that are neither consensus nor EE.

zie1ony:
1. Priority lane for WASM-less transfers allows for a cheap attack that prevents smart contract execution. You just have to flood the network with transfers. 
2. I'm interested in the refund mechanism. On Ethereum it's often impossible to predict the gas usage, because it might depend on the previously executed transactions in the same block. It might cost 1% or 80% gas limit. The best example is an arbitrage contracts, that checks prices on multiple dexes and if there is a arbitrage opportunity, then dexes are called (high gas usage). If there's nothing to win, contract just ends (low gas usage). Those use cases requires setting the refund percentage parameter very low, but that allows cheaply to stop the network by sending a lot of deploys with the high spot gas reservation.

zie1ony (changes requested):
1. Priority lane for WASM-less transfers allows for a cheap attack that prevents smart contract execution. You just have to flood the network with transfers. 
2. I'm interested in the refund mechanism. On Ethereum it's often impossible to predict the gas usage, because it might depend on the previously executed transactions in the same block. It might cost 1% or 80% gas limit. The best example is an arbitrage contracts, that checks prices on multiple dexes and if there is a arbitrage opportunity, then dexes are called (high gas usage). If there's nothing to win, contract just ends (low gas usage). Those use cases requires setting the refund percentage parameter very low, but that allows cheaply to stop the network by sending a lot of deploys with the high spot gas reservation.

AlexanderLimonov:
> 
> 
>     1. Priority lane for WASM-less transfers allows for a cheap attack that prevents smart contract execution. You just have to flood the network with transfers.
> 
>     2. I'm interested in the refund mechanism. On Ethereum it's often impossible to predict the gas usage, because it might depend on the previously executed transactions in the same block. It might cost 1% or 80% gas limit. The best example is an arbitrage contracts, that checks prices on multiple dexes and if there is a arbitrage opportunity, then dexes are called (high gas usage). If there's nothing to win, contract just ends (low gas usage). Those use cases requires setting the refund percentage parameter very low, but that allows cheaply to stop the network by sending a lot of deploys with the high spot gas reservation.

1. You can't flood anything - you're still competing on price with other WASM-less deploys, and, if I actually include the idea for overflowing WASM-less transfers, on price and volume with general deploys.
2. Our intent is to build a platform geared towards integration as corporate middleware, so it's unclear what the value of encouraging DeFi speculation would be in the long run. I would anticipate that such uses would have predictable (in principle) gas requirements, but it's hard to say what the prospective market would really look like. All that's clear is that we cannot have high refunds, because that's both an opening for malicious actors and a disincentive for responsible development of dApps. I would also anticipate that the value of the refund parameter would change over time as use cases are discovered and evaluated.

zie1ony:
> * You can't flood anything - you're still competing on price with other WASM-less deploys, and, if I actually include the idea for overflowing WASM-less transfers, on price and volume with general deploys.

Based on [this paragraph](https://github.com/CasperLabs/ceps/blob/Gas_spot_market/text/0022-gas-spot-market.md#balance-check--allocation) I understood that the strategy is to firstly include all possible transfers. That means if there is more transfers then block can have, no other smart contract call would be executed. Is that correct?

> * Our intent is to build a platform geared towards integration as corporate middleware, so it's unclear what the value of encouraging DeFi speculation would be in the long run. I would anticipate that such uses would have predictable (in principle) gas requirements, but it's hard to say what the prospective market would really look like. All that's clear is that we cannot have high refunds, because that's both an opening for malicious actors and a disincentive for responsible development of dApps. I would also anticipate that the value of the refund parameter would change over time as use cases are discovered and evaluated.

I put DeFi just as an example, there are other. (BTW CasperLabs' official stand is to support DeFi). My point is that the user shouldn't be penalized for no being able to predict future.

text/0022-gas-spot-market.md
```diff
@@ -0,0 +1,215 @@
...
+The limit on spot gas reservations by general deploys.
+
+#### WASM-less transfer dedicated gas limit
+
+The limit on spot gas reservations of WASM-less deploys, naturally stated as a multiple of desired number of such deploys multiplied by their (fixed) gas reservation. 
+
+### Deploy channels
+
+We have already mentioned the concept of a dedicated gas limit for WASM-less transfers in our description of the proposed chainspec parameters. To avoid crowding out WASM-less transfers, which are expected to be critical for platform health and public perception, there should be a separate "lane" for such transfers and this parameter specifies the width of this "lane." Selection among WASM-less transfers would work the same as it does for general deploys, but with sorting by WTP alone, since the gas reservation is fixed for these. Moreover, WASM-less transfers that are not selected for inclusion would be eligible for inclusion as general deploys, with stated WTP and fixed gas reservation.
+
```
dwerner, line 96: From a technical perspective, I think it's best if we keep `transfers` and `wasm_deploys` separate. It's part of an external interface, and we should maintain the separation there (as is desired in https://casperlabs.atlassian.net/browse/NDRS-735).

text/0022-gas-spot-market.md
```diff
@@ -0,0 +1,215 @@
...
+(validator-side, execution)
+
+1. Execute session code as usual, but subject to the available tokens in both the originating account purse and the provided purse.
+
+#### Charges
+
+(validator-side, execution)
+
+1. Charge the provided purse first, charging willingness to pay times gas reservation to the provided purse
```
MParlikar, line 161: I believe we will deduct the payment amount and the motes/gas exchange rate (gas-price) will determine the amount of compute that is actually purchased.

text/0022-gas-spot-market.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+### Gas requirements estimation
+
+We have already mentioned that, due to the consensus-before-execution model, we cannot charge only for gas used during execution. Charging only for gas used fails to provide any incentive for the users
+to estimate gas requirements and open the proposed system to gaming by setting arbitrarily high gas reservations. This is contrary to the notion high-value transactions be prioritized and could possibly result in consistently underfilled blocks. However, this is a serious UX issue, since we have few tools at present for estimating gas requirements. It is expected that tooling will eventually be developed by third parties, using a testnet as a simulacrum of a live network. Such tooling may be developed from existing tests for simple contracts, but would require considerable work to enable testing of complex assemblages of contracts.
+
```
MParlikar, line 172: The cargo-casper infrastructure could be leveraged to consume the opcode costs table.   This would not be a huge lift, and then the result of running a suite of tests, it would be possible to return the total gas used in execution.  The cargo-casper infrastructure could also be used to address complex assemblages of contracts as well.

MParlikar (changes requested):

text/0022-gas-spot-market.md
```diff
@@ -0,0 +1,215 @@
...
+### Basic concepts
+
+#### Willingness to pay (WTP)
+
+Willingness to pay can be seen as analogous to "price" as informally used when discussing user bidding for gas. We will shortly
+address why "price" should not be used for the deploy parameter we are about to define. Willingness to pay is almost self-explanatory, 
+being the amount a user is willing to pay for a unit of gas. We are using the term pragmatically, not intending to mean that this is the
+*maximum* a user is willing to pay. In fact, under our scheme, bidding is almost certainly going to be strategic, with users "shading" their
+true willingness to pay, since the proposed allocation rule is analogous to a first-price auction.
```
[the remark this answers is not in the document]
AlexanderLimonov: I agree that it's rather unfortunate that it clashes with our field names, but if this were implemented, the headline "gas price" as reported at some hypothetical gas station website would be the thing I'm describing. It's not really critical to the substance of the CEP, however.

text/0022-gas-spot-market.md
```diff
@@ -0,0 +1,215 @@
...
+
+#### Spot gas reservation
+
+Spot gas reservation, conceptually, is the claim a user will lay on gas in a proposed block. Similarly to willingness to pay, it will be
+a deploy parameter. The user effectively buys this amount, priced at the stated willingness to pay, when included in a block.
+
+#### Gas allocation
+
+Gas allocation with a greedy revenue-optimizing ordering of deploys is the core activity that this CEP proposes. It is envisioned that this
+rule will be hardcoded into the block proposer component. The rule will be that highest WTP deploys with the highest spot gas reservations
```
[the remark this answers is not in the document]
AlexanderLimonov: That solution may be very far away in the future (year or more), whereas leaving FIFO in place depresses validator revenue and leads to economic inefficiency (someone tinkering around is as likely to get in as someone trying to put through a financially critical transaction). It's also a weak attack vector (with tentative 5000 blocks per week, 10^13 gas limit and price of token at 5 cents, it would cost about $2.5 million to render the platform useless for a week with no op deploys that eat all of the gas).

There's also the slight UI problem of having users specify a gas price that's only ever used for unit conversions. A user paying attention should set this to the minimum of one mote per unit of gas and specify total amount as required to fit the deploy (this behavior is also what causes this to depress validator earnings).

text/0022-gas-spot-market.md
```diff
@@ -0,0 +1,215 @@
...
+    
+    a. Check that originating account purse holds an amount equal to willingness to pay times gas reservation, record the remainder for this originating account purse.
+    
+    b. If sufficient tokens, include the deploy in the protoblock.
+    
+    c. Upon encountering the same originating account purse subsequently, check willingness to pay times gas reservation against remainder, and update the remainder.
+    
+    d. If sufficient tokens in the remainder, include the deploy in the protoblock.
+    
+2. Continue traversing deploy-by-deploy until the gas limit is full.
```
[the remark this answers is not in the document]
AlexanderLimonov: I believe @dwerner had a ticket regarding "respecting block limits" which included these checks.

text/0022-gas-spot-market.md
```diff
@@ -0,0 +1,215 @@
...
+
+3. Once gas limit cannot be filled by the next deploy with an eligible balance or remainder, skip deploys until a deploy small enough is encountered to be considered, otherwise continuing as above.
+
```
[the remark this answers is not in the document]
AlexanderLimonov: I remain confused whether we are _absolutely_ decided on FIFO. The small complexities in this CEP can be removed. The actual substance is just a single sort, instead of grabbing deploys from the buffer as they come (however, for full block gas utilization, we probably do want to skip "too large" deploys _even_ _in_ _FIFO,_ instead of just stopping and sending out a potentially near empty block).

text/0022-gas-spot-market.md
AlexanderLimonov: I'm afraid everyone is getting confused by the described UI, which doesn't actually matter for the critical logic, it just simplifies exposition (or so I thought).

A user can specify a 

(gas price, total payment amount) [what's currently done I think] 
(gas price, total gas reserved) [what this CEP uses, to simplify explanation of the reservation logic]
(total payment amount, total gas reserved)

You can always recover the missing piece from the other two, these all supply the same information, ultimately.

AlexanderLimonov: The one advantage the described UI would have is never having to divide anything, however.

xcthulhu:

PR #23, opened 2020-11-27
Handling nodes sending invalid messages

fizyk20:
Outline of how and when to disconnect from nodes that send us invalid data.

[Rendered](https://github.com/fizyk20/ceps/blob/hwy-109/text/0023-disconnect-from-offenders.md)

### DIFF
--- a/text/0023-disconnect-from-offenders.md
+++ b/text/0023-disconnect-from-offenders.md
@@ -37,7 +37,7 @@ There is also a question of what is the adequate response to an offence for each
 
 Thus, we would define a new mechanism: a `DisconnectAndBlacklist` request to the networking component (see [unresolved question 1](#unresolved-questions)).
 
-The `DisconnectAndBlacklist` request would cause the networking component to disconnect from the given node and add it to a blacklist, preventing it from reconnecting in the future. Specifically, it would blacklist the node's endpoint (IP and port). In order to keep the size of the blacklist manageable, we could consider various kinds of compression mechanisms and simplifications, such as blacklisting a whole IP address (all ports) if the same address shows up multiple times.
+The `DisconnectAndBlacklist` request would cause the networking component to disconnect from the given node and add it to a blacklist, preventing it from reconnecting in the future. Specifically, it would blacklist the node's endpoint (IP and port) - blacklisting a key wouldn't be a solution, since creating a new key is cheap. In order to keep the size of the blacklist manageable, we could consider various kinds of compression mechanisms and simplifications, such as blacklisting a whole IP address (all ports) if the same address shows up multiple times.
 
 ## Reference-level explanation
 
--- a/text/0023-disconnect-from-offenders.md
+++ b/text/0023-disconnect-from-offenders.md
@@ -4,7 +4,7 @@
 
 [summary]: #summary
 
-CEP PR: [casperlabs/ceps#0000](https://github.com/casperlabs/ceps/pull/0000)
+CEP PR: [casperlabs/ceps#0023](https://github.com/casperlabs/ceps/pull/23)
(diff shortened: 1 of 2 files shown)


afck (commented):
My main concern is that in some cases, the remedy could be worse than the disease:
My understanding is that starting a node with a new ky is very easy: The public key that we would blacklist and accuse is extremely cheap. Gossiping an accusation across the network to get everyone to ban a key might be much more expensive than creating a new key and doing it again.

Attackers could try to:
1. cause accusation spam in the network, or
2. make our blacklists in memory grow indefinitely.

So I wonder whether we should actually drop the `Accusation` part entirely, to avoid 1. And for point 2, we could limit the number of entries in the blacklist, and maybe compress it a bit, e.g. as soon as there are 10 different ports blacklisted for the same IP, blacklist the whole IP…

fizyk20:
> The public key that we would blacklist and accuse is extremely cheap

Yes, that's why it definitely shouldn't be the _only_ thing to be blacklisted. An endpoint or an IP address is a better choice there, but the key could also be blacklisted just as an additional countermeasure (why make it easier for the attacker and not do it?).

I see your point, though, and maybe dropping the `Accusation` part is a good idea. If they don't try sending invalid messages to other nodes, there is probably no harm in letting them stay connected, and they would get disconnected as soon as they try that again.

afck:
> the key could also be blacklisted just as an additional countermeasure

Yes, that's totally fine with me, as long as we limit the memory we use for storing blacklisted keys, and we don't _gossip_ those keys, or at least limit the bandwith we allow for gossiping.

Unless we address those concerns, however, I think it would be best to just forget about the keys.

afck (approved):

text/0023-disconnect-from-offenders.md
```diff
@@ -0,0 +1,91 @@
...
+
+By choosing not to accuse nodes sending invalid messages to other nodes, we theoretically make it possible for a known malicious node to participate in the network for a while. However, the risk involved is rather small (see [Rationale and alternatives](#rationale-and-alternatives) below).
+
+## Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+Most of the rationale behind the proposal has been explained [above](#motivation).
+
+One missing part is why not accuse offending nodes. The issue here is that accusation are a nonzero, even if small, cost to the network. A determined malicious party could trick the network into spamming itself with accusations by starting many nodes sending invalid messages. On the other hand, the risk associated with not immediately banning such a node is small. If an offender only sends valid messages to other nodes, there is no harm in still letting it communicate with the network, and if it tries sending an invalid one, it will just be blacklisted by the recipient independently. This means that we wouldn't be gaining much by accusing such nodes, but could potentially introduce a vector for a DoS-type attack.
```
goral09, line 69: You could mention that creating new node identity (keys) is cheap so banning a key is not a solution.

goral09 (approved):

goral09:
Oh, one more thing I remembered: we may actually not need the blacklisting feature. The bigger the network the less we need it.

With Kademlia, the network is not fully connected – every node knows only about a subset of the whole network and that size is also limited. 

For example, the Ethereum node's implementation limits the number of peer to 4 thousand. Kademlia replaces the peers that are unresponsive with the new ones. Plus, there's the rating/scoring system that some of the Ethereum node implementations have, that improves on that algorithm even more. After banning a peer (and disconnecting from it) we would replace it with a new one, hopefully, an honest peer. If that new peer is responsive and honest it will never get replaced by a dishonest ever again. This means, that in a big network with enough honest nodes, everyone will converge towards being connected to only healthy nodes.

PR #24, opened 2020-12-09
Design document for tracking validator set when synchronizing

xcthulhu:
Rendered: https://github.com/xcthulhu/ceps/blob/NDRS-663/text/0024-track-validator-set-when-synchronizing.org


xcthulhu:
This is still a draft because there are two other documents which address outstanding issues I've linked to in my [Unsolved Questions](https://github.com/xcthulhu/ceps/blob/NDRS-663/text/0024-track-validator-set-when-synchronizing.org#unresolved-questions) section.

text/0024-track-validator-set-when-synchronizing.org
```diff
@@ -0,0 +1,141 @@
...
+
+We could add a field =validators= to this with type
+[[https://github.com/xcthulhu/casper-node/blob/0a7f9e5fd7608e2f6574c1e213bd9f5e35880af5/types/src/auction/types.rs#L15][=types::ValidatorWeights=]] (i.e. =BTreeMap<PublicKey, U512>=). This would
+be updated as switch blocks are processed.
+
+To support this, blocks will include the list of validators whenever
+they are updated.  There are two scenarios where this can happen:
+
+  1. In a switch block when new validators become bonded/unbonded
+  2. When a validator has been slashed, although this may not be necessary
```
xcthulhu, line 116: @afck my understanding of the protocol is that this is not necessary...

afck: Yes, I'd actually remove point 2: From the protocol's perspective, slashing does _not_ immediately update the validator set.
(The protocol internally ignores the faulty validator's votes, but that's different.)

text/0024-track-validator-set-when-synchronizing.org
```diff
@@ -0,0 +1,141 @@
...
+As we can see above, a switch block can change the validator
+set. Joining nodes needs to track this.
+
+** Reference-level explanation
+:PROPERTIES:
+:CUSTOM_ID: reference-level-explanation
+:END:
+
+/Synchronization mode/ is mediated by the variant
+[[https://github.com/xcthulhu/casper-node/blob/0a7f9e5fd7608e2f6574c1e213bd9f5e35880af5/node/src/components/linear_chain_sync.rs#L85-L93][=linear_chain_sync::State::SyncingDescendents]]=.
```
afck, line 106: "descendant" (also below)

afck (commented):
Looks good, but I'm wondering whether the switch block should contain the validators of the _next_ era or the ones of the latest era (even further in the future) that it knows. In other words: Should the _switch block_ contain the validator set, or the (much earlier) _booking block_?

The latter would allow us to skip more blocks when syncing. Once we're able to gossip the state, we could skip blocks:
* If the switch block in era 5 contains the validators for era 6, we can directly download the switch block of era 6 next (without any intermediate blocks) and check its signatures.
* If the switch block in era 5 contains the validators for era 10, because it's the booking block for era 10, we could even download the switch block of era 10 next, without caring about anything in between.

xcthulhu:
> 

Looks good, but I'm wondering whether the switch block should contain the validators of the next era or the ones of the latest era (even further in the future) that it knows. In other words: Should the switch block contain the validator set, or the (much earlier) booking block?
>
> The latter would allow us to skip more blocks when syncing. 

I know we want to do fast sync eventually, so having the further future validators is desirable.

However, if I have a trusted hash `h` and I am syncing, how do I know the validators when I hit the next switch block?

afck:
> However, if I have a trusted hash `h` and I am syncing, how do I know the validators when I hit the next switch block?

Yes, it would certainly complicate things. Let's assume `h` is from somewhere in the middle of era 10, the auction delay is 5, and the current era is 22.
* I download block `h`.
* I go backwards from `h`, downloading ancestors until I hit the switch block `s9` of era 9.
* `s9` tells me the validator set of era 14, so I download the switch block `s14` of era 14 and verify its signatures.
* `s14` tells me `s19`, so I do the same for that.
* Now to learn the validator set of the current era 22 I need the switch block from era 17. So I'd have to go backwards from `s19` until I hit that. :confused: 
* `s17` tells me the validators of the current era, so I can download the most recent block and validate its signatures.

It sounds like it would make sense to also add some additional hashes of older blocks, e.g. of the previous switch block, so that we can move back in time faster as well.

fizyk20 (approved):
-- Uhh, got posted under a wrong PR. Sorry! --

goral09:
@afck I'm wondering whether it is _that_ much faster. I guess it would download `1 / AUCTION_DELAY` of blocks? Since it will download every `AUCTION_DELAY`-th switch block. Which, for low values of `AUCTION_DELAY` may not be that much. On the other hand, with short eras it might be a lot...

afck:
Yes, that's true. It's certainly an optimization that could wait until the network is a bit older.
(On the other hand, it would be an upgrade that involves changing the block structure, or at least its semantics.)

text/0024-track-validator-set-when-synchronizing.org
```diff
@@ -0,0 +1,141 @@
...
+:CUSTOM_ID: summary
+:END:
+
+CEP PR: [[https://github.com/casperlabs/ceps/pull/0024][casperlabs/ceps#0024]]
+
+A node joining the network needs a secure means of tracking the
+validator set as it processes blocks. To achieve this, the validators
+for the next era are to be provided in each switch block.
+
+This is required for [[https://github.com/casperlabs/ceps/pull/1234][casperlabs/ceps#1234]] and [[https://github.com/casperlabs/ceps/pull/9999][casperlabs/ceps#9999]].
```
fizyk20, line 19: These are placeholders for some future CEPs, right? I would make it clearer, for example like this:

"This is required for [Feature 1 (placeholder link)](https://github.com/casperlabs/ceps/pull/1234) and [Feature 2 (placeholder link)](https://github.com/casperlabs/ceps/pull/9999)."

Placeholders below could also be replaced with links like that.

xcthulhu:
@henrytill @marc-casperlabs As we've discussed, I'll run `pandoc` on this to convert it to markdown after it's approved and we want to go ahead and merge.

text/0024-track-validator-set-when-synchronizing.org
xcthulhu: Fixed in https://github.com/CasperLabs/ceps/pull/24/commits/351be7f14f7c9a6287a4d3f2e86e8e9f30840e9d

text/0024-track-validator-set-when-synchronizing.org
xcthulhu: Fixed in https://github.com/CasperLabs/ceps/pull/24/commits/953a844d9a6cb0beff74926ef1ea1b87aed99d49

text/0024-track-validator-set-when-synchronizing.org
xcthulhu: Clarified in https://github.com/CasperLabs/ceps/pull/24/commits/9a2daaeb1719b995c9008c82d6045f4eef2cc6b4

xcthulhu:
> In other words: Should the switch block contain the validator set, or the (much earlier) booking block?

I've added a note that we could want future validators in booking blocks in https://github.com/CasperLabs/ceps/pull/24/commits/1d18d8c928a395aac9cdf4ef7a58d62c66824f08

afck (approved):

text/0024-track-validator-set-when-synchronizing.org
```diff
@@ -0,0 +1,142 @@
...
+- How will we enforce the new invariants?  What is the testing strategy?
+
+- What else is needed to support fast sync?
+
+** Future possibilities
+:PROPERTIES:
+:CUSTOM_ID: future-possibilities
+:END:
+
+To support fast synchronization, /booking blocks/ could contain the
```
goral09, line 138: It's worth pointing out that booking block _is_ a switch block – just in/for a different era.

text/0024-track-validator-set-when-synchronizing.org
```diff
@@ -0,0 +1,142 @@
...
+:CUSTOM_ID: unresolved-questions
+:END:
+
+- While validators are now tracked, how will
+  blocks be checked as discussed in the [[#guide-level-explanation][Guide-level
+  explanation]]? This is addressed in [[https://github.com/casperlabs/ceps/pull/1234][casperlabs/ceps#1234]].
+
+- What changes need to be made to transitioning from the
+  =SyncingDescendants= state to ordinary validation?  This is addressed
+  in [[https://github.com/casperlabs/ceps/pull/9999][casperlabs/ceps#9999]].
```
goral09, line 127: Here, and in lines 123, you have a placeholder link.

text/0024-track-validator-set-when-synchronizing.org
```diff
@@ -0,0 +1,142 @@
...
+ 
+- How will we enforce the new invariants?  What is the testing strategy?
+
+- What else is needed to support fast sync?
```
goral09, line 131: Is this a rhetorical question? We know that we need at least the ability to download a global state at specific block.

text/0024-track-validator-set-when-synchronizing.org
```diff
@@ -0,0 +1,142 @@
...
+
+** Motivation
+:PROPERTIES:
+:CUSTOM_ID: motivation
+:END:
+
+As a node gets blocks and becomes up-to-date, it will process blocks
+in ascending eras. If the node has secure knowledge of the validators
+in that era it can use /finality signatures/ to determine the
+authenticity of block hashes. A procedure for tracking the validators
```
goral09, line 30: The authenticity of block hashes or blocks? Is that a mental shortcut?

text/0024-track-validator-set-when-synchronizing.org
```diff
@@ -0,0 +1,142 @@
...
+for each era is essential to securely joining the network.
+
+** Guide-level explanation
+:PROPERTIES:
+:CUSTOM_ID: guide-level-explanation
+:END:
+
+Joining a network requires that a node process one block at a time
+(under the current scheme). It does so until it becomes current. The
+node then runs the /highway/ protocol.
```
goral09, line 40:
```suggestion
node then runs the /Highway/ protocol.
```

text/0024-track-validator-set-when-synchronizing.org
```diff
@@ -0,0 +1,142 @@
...
+As we can see above, a switch block can change the validator
+set. Joining nodes needs to track this.
+
+** Reference-level explanation
+:PROPERTIES:
+:CUSTOM_ID: reference-level-explanation
+:END:
+
+/Synchronization mode/ is mediated by the variant
+[[https://github.com/xcthulhu/casper-node/blob/0a7f9e5fd7608e2f6574c1e213bd9f5e35880af5/node/src/components/linear_chain_sync.rs#L85-L93][=linear_chain_sync::State::SyncingDescendants]]=.
```
goral09, line 107: This renders poorly on GitHub:
image

goral09:
> […] Let's assume `h` is from somewhere in the middle of era 10, the auction delay is 5, and the current era is 22.

I think there are implicit questions here for which we don't know answers for yet:
* How do I know what is the _current_ era? 
* What if my trusted hash is in some earlier era? 
* What if, while I am joining, the current era advances,
* etc.

text/0024-track-validator-set-when-synchronizing.org
goral09: Ah, I see. It's only part of the whole design…

text/0024-track-validator-set-when-synchronizing.org
goral09: OK, I understand now why.

goral09 (approved):
No real blockers, just some nitpicks.

text/0024-track-validator-set-when-synchronizing.org
xcthulhu: I believe I mean block hashes.  Maybe a mental shortcut...?  Not sure what you mean.

text/0024-track-validator-set-when-synchronizing.org
```diff
@@ -0,0 +1,141 @@
...
```
xcthulhu, line 106:
```suggestion
[[https://github.com/xcthulhu/casper-node/blob/0a7f9e5fd7608e2f6574c1e213bd9f5e35880af5/node/src/components/linear_chain_sync.rs#L85-L93][=linear_chain_sync::State::SyncingDescendants=]].
```

text/0024-track-validator-set-when-synchronizing.org
xcthulhu: Should be fixed here: https://github.com/CasperLabs/ceps/pull/24/commits/e440c640aaefb9589df2eee99259e06a0c05c2f6

text/0024-track-validator-set-when-synchronizing.org
xcthulhu: So is every booking block a switch block, but not vice versa?

text/0024-track-validator-set-when-synchronizing.org
xcthulhu: Got rid of these in https://github.com/CasperLabs/ceps/pull/24/commits/17e1433250bcc8b81edd0ecbf53ed8679821e023

text/0024-track-validator-set-when-synchronizing.org
xcthulhu: Got rid of these in https://github.com/CasperLabs/ceps/pull/24/commits/17e1433250bcc8b81edd0ecbf53ed8679821e023

xcthulhu:
> I think there are implicit questions here for which we don't know answers for yet:

Well, I think we can field some of these

>   - How do I know what is the current era?

As the synchronizer pulls down blocks, it pulls down block headers, which have an `era_id` field.

>    What if my trusted hash is in some earlier era?

 When you get the next switch block, the plan is that we'll change the format to contain the validators from the next era: https://github.com/CasperLabs/ceps/pull/24/files#diff-f8e8131af7fa99b018c3f14019a3dccc7f66caefc08d2b8f0b4da60d19802a29R111-R113

text/0024-track-validator-set-when-synchronizing.org
goral09: Booking block and key block are both switch blocks. The difference is that a booking block for era `n` is a switch block from era `n – AUCTION_DELAY` and key block for era `n` is (or `n – m` where `m` is configurable; currently it's `n` ) a switch block of era `n`. 

@afck can you double check me here?

text/0024-track-validator-set-when-synchronizing.org
goral09: The way I was seeing it was that finality signatures are used to determine an authenticity of a block we're downloading. I guess you can say _block hash_ as well since it uniquely identifies a block. Although, for the purposes of the joining process a node wants whole blocks, not only block hashes. That's what I called a mental shortcut.

goral09:
> As the synchronizer pulls down blocks, it pulls down block headers, which have an era_id field.

It doesn't know what is the _current_ era though. I don't think it's important for the example though. I misinterpreted Andreas' comment.

Disregard that comment though. I wrote it before I realized that _this_ CEP is not trying to address all questions regarding the secure joining process but only part of it. They were also rhetorical since we already know answers to them. My point was to note that they're not included in this CEP.

text/0024-track-validator-set-when-synchronizing.org
afck: Agreed, ± possibly off-by-one errors…
(E.g. the key block for era `n` is the switch block from era `n - 1`.)

PR #25, opened 2020-12-09
No Ticket: Introducing template CEP for emacs org-mode

xcthulhu:
Rendered: https://github.com/xcthulhu/ceps/blob/no-ticket-org-mode-template/0000-template.org

### DIFF
--- a/0000-template.org
+++ b/0000-template.org
@@ -1,0 +1,135 @@
+#+STARTUP: inlineimages
+
+* Title
+:PROPERTIES:
+:CUSTOM_ID: title
+:END:
+
+** Summary
+:PROPERTIES:
+:CUSTOM_ID: summary
+:END:
+
+CEP PR:
+[[https://github.com/casperlabs/ceps/pull/0000][casperlabs/ceps#0000]]
+
+One paragraph explanation of the feature.
+
+** Motivation
+:PROPERTIES:
+:CUSTOM_ID: motivation
+:END:
+
+Why are we doing this? What use cases does it support? What is the
+expected outcome?
+
+** Guide-level explanation
+:PROPERTIES:
+:CUSTOM_ID: guide-level-explanation
+:END:
+
+Explain the proposal as if it was already approved and implemented. That
+generally means:
+
+- Introducing new named concepts.
+- Explaining the feature largely in terms of examples.
+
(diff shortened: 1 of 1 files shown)


xcthulhu:
@marc-casperlabs I find it much easier to create diagrams in `org-mode` than markdown.  I've translated the template using pandoc for this reason.  I want to have this template around for myself, but I understand that @fizyk20 also uses emacs as well.

xcthulhu:
Oops!  Looks like @fizyk20 actually uses neovim!  Hopefully this is okay anyway...

fizyk20 (approved):
Even though I'm not an Emacs or org-mode user, I don't see any reason _not_ to add this template - it would be useful at the very least to @xcthulhu , and maybe more people will get interested in what org-mode is.

I'll approve this, but I feel that I'm not the best reviewer for this, so you should probably wait for an opinion from someone else :wink:

marc-casperlabs:
Well there's two things I need to know here:

* Is the end result going to be a Markdown file in the repo? The majority of users (I assume) are not going to be familiar enough with emacs to make use of this, but I'd be fine with it, if, at the end, there is a `.md` in the master branch.
* If we change the template, there's some maintenance burden, but given how rarely we are changing that, I think that's fine?

xcthulhu:
> Is the end result going to be a Markdown file in the repo?

No.  But github renders them exactly the same and the org-mode syntax is super simple.

> If we change the template, there's some maintenance burden, but given how rarely we are changing that, I think that's fine?
I just did `pandoc 0000-template.md -o 0000-template.org` and cleaned a little whitespace.

marc-casperlabs:
> No. But github renders them exactly the same and the org-mode syntax is super simple.

This would essentially change the standard file format from markdown to org-mode, I think. I'm not a fan of that, even though I realized org mode has similar features.

We could support an org-mode workflow by putting the template in a `contrib` folder or something like that --- is it possible to run `pandoc` as a last step to convert the output back to markdown once done? As far as I am concerned, it's mostly the CEP original author that's making changes, so I'm fine with them having their favorite platform supported, but I'd like to preverse future compatibility by having the end product be simple markdown.

xcthulhu:
I chatted with some other folks and it appears we aren't interested in supporting two CEP formats.

PR #26, opened 2020-12-12
NDRS-664: Signatures for Joining Node Synchronization

xcthulhu:


xcthulhu:
Rendered: https://github.com/xcthulhu/ceps/blob/0793fd53ebaa0ef15c07cda4a60f4286f3069e88/text/0026-finality-signature-gossiping.org

goral09:
@xcthulhu Is this CEP ready for a review?

xcthulhu:
> @xcthulhu Is this CEP ready for a review?

Yup!

text/0026-finality-signature-gossiping.org
```diff
@@ -0,0 +1,195 @@
...
+:PROPERTIES:
+:CUSTOM_ID: finality-signatures-for-secure-joining-node-synchronization
+:END:
+
+** Summary
+:PROPERTIES:
+:CUSTOM_ID: summary
+:END:
+
+CEP PR: [[https://github.com/casperlabs/ceps/pull/0025][casperlabs/ceps#0025]]
```
afck, line 13:
```suggestion
CEP PR: [[https://github.com/casperlabs/ceps/pull/0026][casperlabs/ceps#0026]]
```

text/0026-finality-signature-gossiping.org
```diff
@@ -0,0 +1,195 @@
...
+
+Nodes joining the network will use /finality signatures/ to securely
+synchronize with the validators. As a node synchronizes, it will get
+signatures of block hashes via the gossip network until it has > 33%
```
afck, line 17: We should probably use the fault tolerance as configured in the chainspec here.

text/0026-finality-signature-gossiping.org
```diff
@@ -0,0 +1,195 @@
...
+    block_hash: BlockHash,
+    block_height: u64,
+    signature: Signature,
+    public_key: PublicKey,
+}
+#+end_src
+
+The signature would be of the bytes =(block_height, block_hash)=.
+
+A faulty validator must by dealt with when syncing.  A validator could
```
afck, line 158:
```suggestion
A faulty validator must be dealt with when syncing.  A validator could
```

text/0026-finality-signature-gossiping.org
```diff
@@ -0,0 +1,195 @@
...
+This is a /pigeon hole/ argument based on the trust model for highway.
+
+** Unresolved questions
+:PROPERTIES:
+:CUSTOM_ID: unresolved-questions
+:END:
+
+- Where is the enum that needs to be extended so that the gossip
+  protocol to support =fetch_finality_signatures_at_height=?
+- Is it necessary to have completed [[https://casperlabs.atlassian.net/browse/NDRS-667][NDRS-667]] before doing this work?
```
afck, line 195: That's done. :heavy_check_mark:

text/0026-finality-signature-gossiping.org
fizyk20: Here and in all places where 33% or 1/3 is mentioned, right?

fizyk20:
This assumes that we know what the validator set for a given height is, right? Does it mean this depends on CEP #24 , or should we just say we assume we have _a_ mechanism for that, not necessarily the one specified in CEP #24?

Either way, I think it would be good to state this explicitly in this doc.

text/0026-finality-signature-gossiping.org
```diff
@@ -0,0 +1,195 @@
...
+
+Then extend [[https://github.com/xcthulhu/casper-node/blob/0a7f9e5fd7608e2f6574c1e213bd9f5e35880af5/node/src/components/linear_chain_sync.rs#L85-L93][linear_chain_sync::State::SyncingDescendants]] with a field
+=validator_finality_signatures= with type =BTreeMap<PublicKey,
+FaultyOrFetched>=. Fetching validator signatures will update this map.
+New signatures are either added or change old signatures to faulty if
+detected.
+
+To determine the weight of a block hash in this structure: add the
+weight of non-faulty signatures with that hash /plus/ the weight of the
+faulty validators. If this is greater than 1/3 of the validator
```
afck, line 184: I think there are two natural ways to count:
1. The one described here: Wait until F + 1 validators have signed or been proven faulty.
2. The alternative: Wait for N - F signatures by non-faulty (as far as we know) validators.

If the assumption — at most F faulty validators — is correct, both work, and 1 is slightly faster.

If there are ≥ N - F faulty validators, both fail, of course.

But if there are strictly between F and N - F faulty validators, the difference is that 1 risks a _safety_ failure — two observers may consider conflicting blocks as finalized —, whereas 2 only risks a _liveness_ failure: You'll never collect your N - F signatures, and stall.

I know I'm the one who brought up 1; but now I wonder whether we should err on the side of safety?
Anyway, I think both options are okay; whatever you all prefer.

afck (commented):
I'm happy with the approach conceptually, but as discussed, I would prefer if we reorganized the block header and body instead of adding more information to the finality signatures themselves.

text/0026-finality-signature-gossiping.org
xcthulhu: Fixed in https://github.com/CasperLabs/ceps/pull/26/commits/2196a7433bfa90b896e7271442df41760c6a4a73

xcthulhu:
> I'm happy with the approach conceptually, but as discussed, I would prefer if we reorganized the block header and body instead of adding more information to the finality signatures themselves.

I agree.  I will update this CEP accordingly.

text/0026-finality-signature-gossiping.org
```diff
@@ -13,8 +13,9 @@
 CEP PR: [[https://github.com/casperlabs/ceps/pull/0026][casperlabs/ceps#0026]]
 
 Nodes joining the network will use /finality signatures/ to securely
-synchronize with the validators. As a node synchronizes, it will get
-signatures of block hashes via the gossip network until it has > 33%
+synchronize with the validators.  These nodes were identified in As a
```
afck, line 16:
```suggestion
synchronize with the validators.  As a
```
Probably a typo, or missing sentence?

xcthulhu: Typo yeah.

xcthulhu: Fixed in https://github.com/CasperLabs/ceps/pull/26/commits/bf15965deb13ab20a1ebf2b8489e3796232354db

xcthulhu:
> I'm happy with the approach conceptually, but as discussed, I would prefer if we reorganized the block header and body instead of adding more information to the finality signatures themselves.

@afck  Provided that finality signatures are bundled with headers then we have enough information there, so no changes to finality signatures necessary.  I've updated this document so that they will be bundled.

text/0026-finality-signature-gossiping.org
xcthulhu: > If there are ≥ N - F faulty validators, both fail, of course.

Easy enough to have either pattern and switch to the other.

xcthulhu:
> This assumes that we know what the validator set for a given height is, right? Does it mean this depends on CEP #24 , or should we just say we assume we have _a_ mechanism for that, not necessarily the one specified in CEP #24?
> 
> Either way, I think it would be good to state this explicitly in this doc.

I went ahead and documented this in the summary.

afck (approved):

text/0026-finality-signature-gossiping.org
```diff
@@ -0,0 +1,182 @@
...
+:PROPERTIES:
+:CUSTOM_ID: motivation
+:END:
+
+Nodes joining the network need to get blocks from the block-chain in a
+secure manner.  If a block hash has /finality signatures/ from enough
+validators then a node can trust it. Since the security model for
+Highway is tolerant of threshold =t%= (default: 33%) of validators being
+byzantine.  If a node can see that greater than =t%= of the validator
+weight has signed a block-hash or is Byzantine, then that node can
```
goral09, line 32: > If a node can see that greater than `t%` of the validator weight […] is Byzantine

Really? I would expect that if a node sees more than the finality threshold been Byzantine then it wouldn't trust the network at all… Doesn't it mean that there was an equivocation catastrophe and the network has to go through the manual restart?

text/0026-finality-signature-gossiping.org
```diff
@@ -0,0 +1,182 @@
...
+enum FaultyOrFetchedFinalitySignature {
+  /// A pair of finality signatures where:
+  /// - `bad_sig1.public_key == bad_sig2.public_key`
+  /// - `bad_sig1.block_height == bad_sig2.block_height`
+  /// - `bad_sig1.block_hash != bad_sig2.block_hash`
+  Faulty {
+    bad_sig1: FinalitySignature,
+    bad_sig2: FinalitySignature
+  },
+  Fetched(FinalitySignature)
```
goral09, line 158: Maybe `Correct`/`Valid`/`Honest` would match better the `Faulty` state?

goral09 (approved):
No blockers really, just some questions for my own understanding.

text/0026-finality-signature-gossiping.org
xcthulhu: > Really? I would expect that if a node sees more than the finality threshold been Byzantine then it wouldn't trust the network at all… 

Well, @afck discusses this matter here: https://github.com/CasperLabs/ceps/pull/26#discussion_r545829208

While perhaps unintuitive, it doesn't violate the security parameters of the protocol.

> Doesn't it mean that there was an equivocation catastrophe and the network has to go through the manual restart?

Not necessarily no.

text/0026-finality-signature-gossiping.org
xcthulhu: It's not necessarily valid, because another finality signature could be gossiped and it would then be labelled as faulty.

goral09: Until that happens it is valid. If another finality signature invalidates it then we will change enum variant to `Faulty`. Like a state machine.

text/0026-finality-signature-gossiping.org
goral09: If more than `ftt` (by weight) validators equivocate that is an equivocation catastrophe though.

If there is an equivocation but weight of equivocators is less than `ftt` – it's not.

In this case:
> If a node can see that greater than t% of the validator weight […] is Byzantine

it is a catastrophe. As per my undertanding.

text/0026-finality-signature-gossiping.org
xcthulhu: Less confusing for me to leave it as `Fetched`.  Does that not work for you?

text/0026-finality-signature-gossiping.org
xcthulhu: We may be talking past one another.

> > If a node can see that greater than t% of the validator weight […] is Byzantine

The full sentence is this:

> If a node can see that greater than =t%= of the validator
> weight has signed a block-hash or is Byzantine, then that node can
> trust that block hash.

You appear to be concerned about the case where more than =t%= of nodes are Byzantine.  In that case the security assumption is violated.

Are you trying to suggest that in that case is detected, the validator node should halt?

goral09: Yes, I do. A validating node will stop and rely on the node operators/validators to agree on the Genesis and restart the network, as described in CEP #6. I think that a similar behaviour should be implemented for a joining node.

text/0026-finality-signature-gossiping.org
xcthulhu: To clarify, it looks like we already have an enum `Observation` [in panorama][1] that is similar, so we should follow that convention.  There we used the term `Correct`.

[1]: https://github.com/goral09/casper-node-1/blob/0bbefb1c284784f0307661cdc7c33584ffded3ce/node/src/components/consensus/highway_core/state/panorama.rs#L24-L31

text/0026-finality-signature-gossiping.org
xcthulhu: I agree that a joining node should fail fast here.

xcthulhu: Updated in https://github.com/CasperLabs/ceps/pull/26/commits/927177407952d5d24b2a245fc9030ead7eb71295

text/0026-finality-signature-gossiping.org
xcthulhu: @afck after talking this over we decided we will try to detect if more than `t%` are Byzantine and stop the process if there is an equivocation catastrophe.  Did this here: https://github.com/CasperLabs/ceps/pull/26/commits/927177407952d5d24b2a245fc9030ead7eb71295

text/0026-finality-signature-gossiping.org
xcthulhu: Fixed here: https://github.com/CasperLabs/ceps/pull/26/commits/cc33b7063751e917c68c428feec34be1d3661ab0

goral09 (approved):

PR #27, opened 2020-12-22
CEP: Disabling Bids on Liveness Failures

fizyk20:
Rendered: https://github.com/fizyk20/ceps/blob/disabling-bids/text/0027-disabling-bids.md

### DIFF
--- a/text/0027-disabling-bids.md
+++ b/text/0027-disabling-bids.md
@@ -39,7 +39,7 @@ In either case, we would want to disable such a validator's bid. This would be d
 
 The validity of this transaction would be determined by the consensus state as seen by the switch block proposal consensus unit (i.e. the number of messages sent by each validator would be calculated based on what the unit containing the switch block sees) - this ensures that all validators will agree about whose bids need to be disabled.
 
-The new transaction would completely disable the validator's bid upon execution by the auction contract and remove the offending validator from the validator set from the next era onwards. The funds would not be redeemable until the regular delay (`LOCKED_FUNDS_PERIOD`) passes. If the validator desires to become a validator again, it would have to submit a new, separate bid.
+The new transaction would completely disable the validator's bid upon execution by the auction contract, which would set a flag on the bid, marking it as inactive, and remove the validator from the eras in which it had already won the auctions. The bid would remain in this state until the validator either bids again (possibly a zero amount, which would remove the "inactivity" flag and make it eligible for winning future auctions again) or explicitly retracts the bid.
 
 ## Reference-level explanation
 
--- a/text/0027-disabling-bids.md
+++ b/text/0027-disabling-bids.md
@@ -35,83 +35,30 @@ Parameters `k`, `N` and `m` should be chosen in a way that achieves the best bal
 
 An inactive validator is probably one that completely lost the connection to the network. A failing validator either experiences intermittent network problems, or is deliberately trying to disrupt the network by sending messages erratically.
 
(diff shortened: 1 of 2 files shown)


AlexanderLimonov (changes requested):
I may be mistaken, but presently we have no real means of "giving back" the tokens unless they're explicitly withdrawn, which should create a purse that the user knows (@EdHastingsCasperLabs). So this proposal for forced unbonding seems unworkable, given my present level of understanding. What I would suggest is that this be re-written to say that signal to the auction contract merely sets a flag on a particular bid, which takes it out of future eras where it's already a winner and disables it for auction runs. Then, the validator can "jiggle" it by adding tokens (possibly 0) and resume bidding.

afck:
I don't think we need a special transaction for this that needs to be gossiped. I'd make it a system transaction, just like slashing and rewards, that depends on what can be seen from the current block. (Maybe only in the switch block, together with slashing and rewards?)

text/0027-disabling-bids.md
```diff
@@ -39,7 +39,7 @@ In either case, we would want to disable such a validator's bid. This would be d
 
 The validity of this transaction would be determined by the consensus state as seen by the switch block proposal consensus unit (i.e. the number of messages sent by each validator would be calculated based on what the unit containing the switch block sees) - this ensures that all validators will agree about whose bids need to be disabled.
 
-The new transaction would completely disable the validator's bid upon execution by the auction contract and remove the offending validator from the validator set from the next era onwards. The funds would not be redeemable until the regular delay (`LOCKED_FUNDS_PERIOD`) passes. If the validator desires to become a validator again, it would have to submit a new, separate bid.
+The new transaction would completely disable the validator's bid upon execution by the auction contract, which would set a flag on the bid, marking it as inactive, and remove the validator from the eras in which it had already won the auctions. The bid would remain in this state until the validator either bids again (possibly a zero amount, which would remove the "inactivity" flag and make it eligible for winning future auctions again) or explicitly retracts the bid.
```
afck, line 42: So that would change eras that had already been decided, and whose random seed is already known? Meaning that if you control N validators, you could choose one out of 2<sup>N</sup> leader sequences without even getting slashed, just by being inactive and getting your bids flagged.

I'd prefer to err on the safe side and only remove them from the validator set after the bonding period.

(We do currently choose the random bit only in the last block before the era… but even that block could be enough, if the attacker creates it: Then they can still choose from which of their validators to include a unit.)

fizyk20: So basically wait for `AUCTION_DELAY` eras after the bid is disabled? Or was there another constant for the bonding period?

afck: You're right, that would be `AUCTION_DELAY`.

afck (approved):

goral09 (approved):

AlexanderLimonov:
Please note that this still should be reviewed by Ed or Henry, and I don't think I can approve the changes I myself suggested, since I was wrong! We have to revert to the original, I believe.

MParlikar (approved):

MParlikar:
Let's get this CEP Merged.  We need to start implementation.

xcthulhu (approved):

fizyk20:
Per Medha's request, I'll merge this as is. If @EdHastingsCasperLabs or @henrytill have any reservations, please state them, and I'll apply necessary changes in a separate PR.

PR #28, opened 2020-12-27
Small auction ce ps

AlexanderLimonov:
First draft


AlexanderLimonov:
Forgot that we need numbers for these

PR #29, opened 2020-12-27
Delegation cap

AlexanderLimonov:
Introducing a delegation cap

[Rendered](https://github.com/CasperLabs/ceps/blob/Delegation/text/0029-delegation-cap.md)


text/0029-delegation-cap.md
```diff
@@ -0,0 +1,67 @@
...
+
+#### Delegation cap multiplier
+
+As described above, this parameter, set to a value always weakly greater than 1, would determine the allowable ratio of total tokens to validator's tokens in a bid. Assuming *x >= 1* is our parameter, *y* is the quantity delegated tokens and *z* is the quantity of validator's tokens, the relationship between the three variables is described as follows,
+
+*z + y <= z \* x*
+
+### Changes to auction and related system contracts
+
+If the validator adjusts his bid in a way that causes the bid to violate the limit, then the weight of the bid (for purposes of protocol leader selection and auction operation) is capped by the multiplier applied to own tokens. Any rewards, however, are to be distributed between the validator and the delegator in accordance to the true quantities of tokens staked.
```
MParlikar, line 41: While this sounds simple, the implementation of this logic is likely going to be very complex, resulting in changes to the rewards mechanism.  It will also make calculation of rewards for delegators harder to predict.

text/0029-delegation-cap.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+The expected drawback to this proposal is the exclusion of high delegation ratio validators, which may make it harder for less risk-averse delegators to achieve desired risk exposure. For instance, a highly performant, unknown and token-poor validating node could be excluded from the market, thereby denying potential delegators outsized returns from delegating to such a node.
+
+## Rationale and alternatives
+
```
MParlikar, line 50: What about exposing the amount of stake that each validator has put up themselves?  And also supporting a 'node name' -  (perhaps done via dns name services) -  and then  exposing this information for delegators to look at.  Selecting a validator that has 'nothing at stake' is an obvious bad choice.  The community could create a staking interface that displays this information.   Then stakers can make informed decisions.

MParlikar (changes requested):

text/0029-delegation-cap.md
```diff
@@ -0,0 +1,67 @@
...
+## Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+We propose to limit the amount of delegated tokens contributing to a bid to a multiplier X of validator’s own token. For example, if a prospective validator’s standing bid is 100 tokens and X is set to 1.2, up to 20 tokens may be delegated. Should a bid be decreased to the point where the quantity of existing delegated tokens violates this constraint, delegated tokens above the limits, in aggregate, should no longer “count” for auction result determination or reward weight.
```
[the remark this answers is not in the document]
AlexanderLimonov: There is only downside to _validators_ here. There are two cases, either the limit is respected (in which case all tokens behave normally), or, else, the limit is breached, in which case delegator tokens are, in a very weak sense, more valuable than validator tokens (since the delegators start eating into validator rewards).

text/0029-delegation-cap.md
AlexanderLimonov: There is no change to the reward mechanism on the consensus side here, because it just considers the eligible blob of tokens (validator and delegator tokens lumped together), only in the system contracts, I believe.

Delegator rewards would be dependent on the ratio on matter what, so anyone with a non-trivial delegation has to monitor the state of the bids they're participating in. This is a UI issue that's solvable with a dashboard (e.g., an expanded version of the thing that got built for bids in delta).

text/0029-delegation-cap.md
AlexanderLimonov: Yes, I think a lot of the complexity of this would go away with a well designed dashboard that lays this information out clearly, together with history. 

I made a CEP that no one ever looked at, regarding this, although it needs more work (simplifying it, for one) - https://github.com/CasperLabs/ceps/blob/Clarity-auction-UI-UX-requirements/text/0013-Clarity-auction-UI-UX-requirements.md

xcthulhu:

PR #30, opened 2020-12-27
Validator lock-up

AlexanderLimonov:
Introducing changes to existing validator lock-up functionality (plus special withdrawal deploy type)

[Rendered](https://github.com/CasperLabs/ceps/blob/Validator_lock_in/text/0030-genesis-validator-lockup.md)

### DIFF
--- a/text/0030-genesis-validator-lockup.md
+++ b/text/0030-genesis-validator-lockup.md
@@ -4,7 +4,7 @@
 
 [summary]: #summary
 
-CEP PR: [casperlabs/ceps#0000](https://github.com/casperlabs/ceps/pull/0000)
+CEP PR: [casperlabs/ceps#0030](https://github.com/casperlabs/ceps/pull/0030)
 
 We propose that the operation of the auction and related contracts be modified to reflect the language of agreements regarding token allocations. To this end, we propose that the unlock of tokens be gradual and take place over the course of 90 days. Further, we introduce a new type of deploy, the genesis validator token withdrawal, which would enable genesis validators to withdraw eligible tokens without paying for gas.
 
--- a/text/0030-genesis-validator-lockup.md
+++ b/text/0030-genesis-validator-lockup.md
@@ -24,7 +24,7 @@ We introduce two changes to the platform. The first change concerns the lock-in 
 - After 90 days, tokens unlock gradually (equal proportion of starting amount) over the course of a further 90 days, possibly as weekly lump sums
 - Auto-win conditions is set to false at the 90 day mark and genesis validators begin to compete in the auction
 
-The second change concerns the problem of genesis validators having no tokens to withdraw their rewards. To this end, we propose introducing a new deploy type, the genesis validator token withdrawal, which requires no payment for gas and can be processed by the platform for the first **210 days**.
+The second change concerns the problem of genesis validators having no tokens to withdraw their rewards. To this end, we propose introducing a new deploy type, the genesis validator token withdrawal, which requires no payment for gas and can be processed by the platform for the first **180 days**.
 
 ## Reference-level explanation
 
--- a/text/0000-genesis-validator-lockup.md
+++ b/text/0000-genesis-validator-lockup.md
@@ -1,0 +1,57 @@
+# Lock-up of genesis validators and access to tokens staked at genesis
+
+## Summary
+
+[summary]: #summary
+
+CEP PR: [casperlabs/ceps#0000](https://github.com/casperlabs/ceps/pull/0000)
+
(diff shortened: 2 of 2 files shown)


text/0030-genesis-validator-lockup.md
```diff
@@ -0,0 +1,57 @@
...
+
+[guide-level-explanation]: #guide-level-explanation
+
+We introduce two changes to the platform. The first change concerns the lock-in period and auto-win conditions managed by the auction contract:
+
+- Total lock-in for 90 days remains in place
+- After 90 days, tokens unlock gradually (equal proportion of starting amount) over the course of a further 90 days, possibly as weekly lump sums
+- Auto-win conditions is set to false at the 90 day mark and genesis validators begin to compete in the auction
+
+The second change concerns the problem of genesis validators having no tokens to withdraw their rewards. To this end, we propose introducing a new deploy type, the genesis validator token withdrawal, which requires no payment for gas and can be processed by the platform for the first **210 days**.
```
goral09, line 27: We expect people to stake millions of dollars and not expect them to buy enough tokens to make the withdrawal (which should be a couple of cents)?

AlexanderLimonov: There are no tokens to buy at the beginning of time, since the initial validators' tokens are locked into stakes. The first liquidity comes from them withdrawing their rewards and unlocked portions of the stake.

goral09: There are people, other than validators, whose public keys are in the initial `accounts.csv` file – they don't have any tokens staked but they had bought tokens on pre-sales and their initial balance is non-zero.

momipsl: The account.csv could be updated so that for each genesis validator, the value in the initial balance column is enough to initiate withdrawal and the initial staked amount is reduced accordingly.

AlexanderLimonov: @goral09, do you mean genesis delegators, perhaps? To date, all information I saw suggested that having to stake tokens "circulating" at genesis was a contractual obligation that actually made the sales possible. It's true that there is a source of liquidity from post-launch sales to non-validators, of course, but it's not ideal to have validators wait for those before being able to access rewards. (sorry, I meant to reply earlier, but events intervened)

@momipsl, I think I get the gist of what you're suggesting, but I thought having genesis validators get unbonded tokens was contrary to the token sale agreements

text/0030-genesis-validator-lockup.md
```diff
@@ -0,0 +1,57 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+We introduce two changes to the platform. The first change concerns the lock-in period and auto-win conditions managed by the auction contract:
+
+- Total lock-in for 90 days remains in place
+- After 90 days, tokens unlock gradually (equal proportion of starting amount) over the course of a further 90 days, possibly as weekly lump sums
+- Auto-win conditions is set to false at the 90 day mark and genesis validators begin to compete in the auction
+
+The second change concerns the problem of genesis validators having no tokens to withdraw their rewards. To this end, we propose introducing a new deploy type, the genesis validator token withdrawal, which requires no payment for gas and can be processed by the platform for the first **180 days**.
+
```
MParlikar, line 28: For VFTA holders (US specifically, but doesn't matter here) - rewards are also locked for 90 days.  Therefore, for all accounts in the genesis block, **no tokens can be withdrawn and rewards cannot be re-bonded for the first 90 days.**

MParlikar (changes requested):

text/0030-genesis-validator-lockup.md
AlexanderLimonov: Ok, so, then, _everyone's_ rewards will be locked, US or non-US, for 90 days, then released all at once?

text/0030-genesis-validator-lockup.md
```diff
@@ -0,0 +1,57 @@
+# Lock-up of genesis validators and access to tokens staked at genesis
+
+## Summary
+
+[summary]: #summary
+
+CEP PR: [casperlabs/ceps#0030](https://github.com/casperlabs/ceps/pull/0030)
+
+We propose that the operation of the auction and related contracts be modified to reflect the language of agreements regarding token allocations. To this end, we propose that the unlock of tokens be gradual and take place over the course of 90 days. Further, we introduce a new type of deploy, the genesis validator token withdrawal, which would enable genesis validators to withdraw eligible tokens without paying for gas.
```
EdHastingsCasperLabs, line 9: We are not adding host side withdrawal.

text/0030-genesis-validator-lockup.md
EdHastingsCasperLabs: That would certainly be the technically easiest to implement.

xcthulhu:

PR #31, opened 2020-12-27
Designated node operators

AlexanderLimonov:
Introducing node operators

[Rendered](https://github.com/CasperLabs/ceps/blob/Designated_operators/text/0031-designated-operators.md)


text/0031-designated-operators.md
```diff
@@ -0,0 +1,51 @@
...
+
+[guide-level-explanation]: #guide-level-explanation
+
+Add an additional field, expecting a public key (or some type wrapping a public key), to the `add_bid` entry point of the auction contract. The supplied public key will be used to sign consensus messages on the validator's behalf and to sign for or identify the validator in all instances required by the protocol proper. The link between the validator and the operator, as well as the history of validator's operator public keys, is to be tracked by the auction contract. Tracking of the pairings between a validator and its operators is intended to enable slashing, distribution of rewards and collection of transaction fees.
+
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+N/a
```
xcthulhu, line 27:
```suggestion
N/A
```

But perhaps instead of fixing the typo, just delete the section?  I haven't been including them because it's just noise.

AlexanderLimonov: Well, Not applicable rather than Not Applicable. I've included them just as a reminder for myself and readers to maintain discipline with these things (and invite people to critique the absence of content).

Fraser999:
I'm not quite following the main gist of the proposal here I'm afraid.

For a validator to be able to sign anything as the operator, it'd need to have access to the operator's secret key, not just its public one.  If the validator has access to the operator's secret key, then I don't see why the validator wouldn't just run using that key - it wouldn't need a keypair of its own.  So I feel I must be missing something fundamental here.

AlexanderLimonov:
> 
> 
> I'm not quite following the main gist of the proposal here I'm afraid.
> 
> For a validator to be able to sign anything as the operator, it'd need to have access to the operator's secret key, not just its public one. If the validator has access to the operator's secret key, then I don't see why the validator wouldn't just run using that key - it wouldn't need a keypair of its own. So I feel I must be missing something fundamental here.

No, the operator does the operating, signing with own private key. The validator points to the operator's public key and says "this is my operator." Otherwise everything works as it does now.

I just realized that the phrasing is incorrect, so I'll fix that.

Fraser999:
So, let me see if I have it right: the idea is that we have a validator `V` and an operator `O`.  `O` runs the node software, passing its own secret key on startup via the config option as per current master.  At some stage `O` sends a deploy calling on the auction contract to add `V`'s public key to this new mapping proposed in this CEP, so that rewards, slashing, etc. are shared between `V` and `O`. Is that correct?

AlexanderLimonov:
> 
> 
> So, let me see if I have it right: the idea is that we have a validator `V` and an operator `O`. `O` runs the node software, passing its own secret key on startup via the config option as per current master. At some stage `O` sends a deploy calling on the auction contract to add `V`'s public key to this new mapping proposed in this CEP, so that rewards, slashing, etc. are shared between `V` and `O`. Is that correct?

Well, not quite. When the bid is created, the validator may choose to specify an operator, by passing in the operator's public key. From then on, operator creates all messages, signing with own private key. However, the validator collects all of the rewards (and the slashes!) as usual. It is implicitly assumed that they have some kind of an off-chain contractual relationship that provides for operator's compensation. Note that there is no *confirmation* for this relationship, as @EdHastingsCasperLabs correctly suggested that this would require a lot of messy code. However, I haven't invented a way in which this could be a problem, so it's fine, provisionally.

Fraser999:
> When the bid is created,

So, on the run up to this point in time, is the validator running their own instance of our casper-node software?  If not, how did they become a validator?  (I guess they could be listed as a genesis validator, but that'll not be the normal case, right?)

> the validator may choose to specify an operator, by passing in the operator's public key.

That's done by the validator sending a deploy, right?

> From then on, operator creates all messages, signing with own private key.

So, at that point, the validator can stop their node from running, and consensus expects all further messages representing this validator to be signed by its operator.

So, I have a couple more questions:
1. At which exact point does this transition happen?
1. Would it be deemed an error if a validator continued running and taking part in consensus while it also has an operator specified?  Seems like it might be easier for the users if we don't treat that as an error, but rather just prefer validator messages over operator ones.  That way, if there's a bit of an overlap while a validator adds or removes an operator, it won't cause a problem.  Not sure how practicable that is to implement from a consensus point of view though.

AlexanderLimonov:
> 
> 
> > When the bid is created,
> 
> So, on the run up to this point in time, is the validator running their own instance of our casper-node software? If not, how did they become a validator? (I guess they could be listed as a genesis validator, but that'll not be the normal case, right?)
> 

Three cases here -

Genesis -> Operator specified at platform launch
Entrant (no history as bidder or validator) -> Optionally supplies operator public key in add_bid
Existing bid -> Changes bid data (by using add_bid, just as they do to increase the bid amount) to include operator public key, platform will expect this operator to sign messages on validator's behalf in a future era where this bid won

Everyone is either a validator at genesis, or enters through winning the auction.

> > the validator may choose to specify an operator, by passing in the operator's public key.
> 
> That's done by the validator sending a deploy, right?

Yes, it's a deploy that calls add_bid.

> 
> > From then on, operator creates all messages, signing with own private key.
> 
> So, at that point, the validator can stop their node from running, and consensus expects all further messages representing this validator to be signed by its operator.

Ideally, the validator going this route shouldn't ever start a node, but the platform should just ignore messages signed by the validator (if any appear) and only look at the operator, instead.

> 
> So, I have a couple more questions:
> 
>     1. At which exact point does this transition happen?

A typical timeline would be something like this

Bid B entered by a new entrant, no operator provided
...
Auction for era N+auction_delay happens, bid B wins
...
Era N+auction_delay starts, B is a validator, signing own messages
During era N+auction_delay, B calls add_bid to supply an operator (and possibly add/reduce number of tokens in the bid)
Auction for era N+auction_delay+auction_delay happens, B wins again
...
Era N+auction_delay+auction_delay starts, B is a validator, but messages are produced and signed by the operator, B-signed messages are ignored

> 
>     2. Would it be deemed an error if a validator continued running and taking part in consensus while it also has an operator specified?  Seems like it might be easier for the users if we don't treat that as an error, but rather just prefer validator messages over operator ones.  That way, if there's a bit of an overlap while a validator adds or removes an operator, it won't cause a problem.  Not sure how practicable that is to implement from a consensus point of view though.

As I said above, if the validator is supposed to be signed for by an operator in a particular era, own-signed messages should just be discarded (as if they are coming from some random node that isn't a validator at all). Otherwise it seems like it's an easy route to enabling accidental equivocations (validator forgets to turn off node, doesn't cite operator's message).

AlexanderLimonov:
@Fraser999, I'll try to add the discussion here to the body of the CEP tomorrow, plus, I think there is another thing we need to deal with - we need to prevent malicious operator designations. This isn't something I considered before, unfortunately. The problem is this - I'm not sure how to reason about a situation where an active validator has designated either _another validator_ or another validator's _operator_ as his own. For simplicity, we don't do confirmations, the designation only requires action by the validator.  

I think this is a check that should run at each auction, and any bid doing the following should be ignored:
- Bid designates another auction participant as operator
- Bid designates another auction participant's operator as an operator

@afck, I think the above should also check against relevant (slashable) validators/operators from past eras?

afck:
I don't understand why we need to distinguish between a "validator with a separate operator" and a delegator. To my mind, a validator who doesn't operate a node _is_ a delegator. From the system's point of view, there aren't three concepts (operator, validator, delegator), but only two:
* The key that signs the consensus messages (operator/validator key).
* The keys that are allowed to unbond (some of) the stake: the delegators.

Sure, in practice, one of the delegators will usually be the same person/organization that actually operates the node, but I don't think the system needs to care about that.

afck:
Thanks for the explanation, @AlexanderLimonov: So the validator key will be the one that has the permisson to _change_ the operator keys (without anyone bonding or unbonding).

afck:
> I think the above should also check against relevant (slashable) validators/operators from past eras?

I agree. But I hope all of this can be implemented purely on-chain, without increasing the surface between consensus and the contract.

EdHastingsCasperLabs (approved):

text/0031-designated-operators.md
```diff
@@ -0,0 +1,86 @@
...
+### Restrictions on designation
+
+In order to simplify reasoning about validator-operator pair behavior and to prevent potential introduction of equivocation or reward bugs, operator designation should be checked against the following:
+
+1. Validator public keys slated to participate in future eras, other than that of the designator.
+2. Validator public keys that are participating in current era or have participated in past eras, up to the unbonding wait period, other than that of the designator.
+3. Bidder public keys for the upcoming auction, other than that of the designator.
+3. Operator public keys corresponding to validator public keys in categories 1 through 3.
+
+Should a designating deploy specify a "duplicate" public contained in the categories 1 through 4, the deploy ought to be rejected. Note that we allow for the possibility of designating oneself as the operator, with no changes to current node or consensus behavior.
```
Fraser999, line 62:
```suggestion
Should a designating deploy specify a "duplicate" public key contained in the categories 1 through 4, the deploy ought to be rejected. Note that we allow for the possibility of designating oneself as the operator, with no changes to current node or consensus behavior.
```

text/0031-designated-operators.md
```diff
@@ -0,0 +1,86 @@
...
+
+Existing bid - changes bid data (by using `add_bid`, similar to adjusting bid token amount) to include an operator public key; platform will expect this operator to sign messages on validator's behalf in a future era where this bid wins
+
+### Typical timeline
+
+Bid entered by a new entrant B, no operator provided
+... (some blocks are added to the linear chain)
+Auction for era N + `auction_delay` takes place, B wins
+...
+Era N+auction_delay starts, B is a validator, signing own consensus messages with own private key
```
Fraser999, line 47:
```suggestion
Era N + `auction_delay` starts, B is a validator, signing own consensus messages with own private key
```

Fraser999 (approved):
A couple of nitpicks, but otherwise LGTM.

text/0031-designated-operators.md
```diff
@@ -18,13 +18,46 @@ It is expected that many validators will want to rely on third parties to provis
...
+At bid creation, the prospective validator may choose to specify an operator, by passing in the operator's public key. If the bid wins the auction, the operator will be responsible for creating all messages, signing with own private key. However, the validator will collect all of the rewards and would be subject to slashing, as usual. Our assumptions is that an off-chain contractual relationship provides for the operator's compensation. Existing validators and bidders are able to switch operators at will, with the change taking effect in the first future era where the new operator is specified in a validator's winning bid. There is no confirmation for the relationship, but provisions have to be made to restrict designation to simplify reasoning about the behavior of the validator-operator pair.
```
afck, line 21: "assumptions is" → "assumption is"

> switch operators at will, with the change taking effect in the first future era where the new operator is specified in a validator's winning bid

I just wanted to say that:

> ~Actually, this is probably something that could even take effect sooner, right after the next era change? It wouldn't change the validator weights, after all.~

But currently the leader schedule also depends on the _order_ of validator keys (lexicographically), which _would_ change if consensus only knows about the operator keys. :confused:

afck: But we could make the consensus "key" just the tuple of both keys, with the validator key first and the operator key second. For cryptography it would only use the operator key, but the ordering would be determined by the validator key.

On the other hand, it's probably surprising behavior if `add_bid` takes effect partly in the next era and partly `auction_delay` eras in the future.

text/0031-designated-operators.md
```diff
@@ -18,13 +18,46 @@ It is expected that many validators will want to rely on third parties to provis
...
+
+### Typical timeline
+
+Bid entered by a new entrant B, no operator provided
+... (some blocks are added to the linear chain)
+Auction for era N + `auction_delay` takes place, B wins
+...
+Era N+auction_delay starts, B is a validator, signing own consensus messages with own private key
+During era N + `auction_delay`, B calls `add_bid` to supply an operator public key (and possibly add/reduce number of tokens in the bid)
+Auction for era N + `auction_delay` + `auction_delay` takes place, B wins again
```
afck, line 54: I think it's era N + `auction_delay` + `auction_delay` + 1. (Or "+ 2", if the first `add_bid` call is made in era N. Unless I'm making another off-by-one error here. :sweat_smile:)

text/0031-designated-operators.md
```diff
@@ -0,0 +1,93 @@
...
+### Restrictions on designation
+
+In order to simplify reasoning about validator-operator pair behavior and to prevent potential introduction of equivocation or reward bugs, operator designation should be checked against the following:
+
+1. Validator public keys slated to participate in future eras, other than that of the designator.
+2. Validator public keys that are participating in current era or have participated in past eras, up to the unbonding wait period, other than that of the designator.
+3. Bidder public keys for the upcoming auction, other than that of the designator.
+3. Operator public keys corresponding to validator public keys in categories 1 through 3.
+
+Should a designating deploy specify a "duplicate" public key contained in the categories 1 through 4, the deploy ought to be rejected. Note that we allow for the possibility of designating oneself as the operator, with no changes to current node or consensus behavior.
```
MParlikar, line 69: Execution of the transaction has to fail (versus deploy being rejected)

MParlikar (approved):

text/0031-designated-operators.md
AlexanderLimonov: Will change

text/0031-designated-operators.md
AlexanderLimonov: Currently consensus doesn't know anything about bid state until the era where new winners take over begins, right? I am not suggesting any changes to that. If you want a new operator, you have to wait a while for it to make a difference in consensus, just as with changing the amount of tokens you bid.

ashok-cl (approved):

PR #32, opened 2021-01-08
NDRS-697: add upgrade-bootstrapper CEP

Fraser999:
[Rendered](https://github.com/Fraser999/ceps/blob/NDRS-697-upgrade-boostrapper/text/0032-upgrade-bootstrapper.md).

### DIFF
--- a/text/0032-upgrade-bootstrapper.md
+++ b/text/0032-upgrade-bootstrapper.md
@@ -22,31 +22,28 @@ Under normal circumstances, the following steps will be performed:
 
 1. the `upgrade-bootstrapper` is started
 2. if the path to the node's `config.toml` was provided via an arg, the `upgrade-bootstrapper` writes the path value to
-   a hard-coded, known location (if writing fails, it panics).  Otherwise it reads the path value from the known
-   location.  If the path cannot be read from the known location, it panics
-3. if in the same folder as the `config.toml` there is a `config-old.toml`, the `upgrade-bootstrapper` runs the
+   a hard-coded, known location (if writing fails, it exits with an error).  Otherwise it reads the path value from the
+   known location.  If the path cannot be read from the known location, it exits with an error
+3. if in the same folder as the `config.toml` there is a `config-next.toml`, the `upgrade-bootstrapper` runs the
    `casper-node` in migrate-data mode (see below for further details), passing it the path to `config.toml` and
-   `config-old.toml` as args.  The node performs any migration of stored data and config options required, then exits
-   normally with an exit code 0.  The node should delete `config-old.toml`, but only once the full migration of data and
-   config is complete.
-4. the `upgrade-bootstrapper` then runs the `casper-node` binary in validator mode, passing it the path to its
-   `config.toml` as an arg
-5. while they are running, before the network reaches the upgrade point, the next version of the `casper-node` gets
-   installed, but not run.  The binary is always named `casper-node-next` and its config named `config-next.toml`.  Each
-   gets installed to the same location as the current equivalent files.  The location of the `config.toml` can be
-   retrieved from the known location used by the `upgrade-bootstrapper` at step 2.  An upgrade file for chainspec
-   modifications should also be installed and its path included in the new `config-next.toml`
-6. when the current upgrade activation point is reached (see [the first unresolved question](#unresolved-questions)
-   regarding how the running node will get to know of the activation point) and the node has finished all its work in
-   that era, it exits normally with an exit code 0
-7. the `upgrade-bootstrapper` renames
-     * `casper-node-next` to `casper-node` (replaces the old binary)
-     * `config.toml` to `config-old.toml`
-     * `config-next.toml` to `config.toml`
-8. repeat from step 3
+   `config-next.toml` as args.  The node performs any migration of stored data, then exits normally with an exit code 0.
+4. the `upgrade-bootstrapper` renames `config-next.toml` if it exists to `config.toml` (replaces the old node config)
+5. the `upgrade-bootstrapper` runs the `casper-node` binary in validator mode, passing it the path to its `config.toml`
+   as an arg
+6. while they are running, before the network reaches the upgrade point, the next version of the `casper-node` gets
+   installed, but not automatically run.  The binary is always named `casper-node-next` and its config named
+   `config-next.toml`.  Each gets installed to the same location as the current equivalent files.  The location of the
+   `config.toml` can be retrieved from the known location used by the `upgrade-bootstrapper` at step 2.  An upgrade file
+   for chainspec modifications should also be installed and its path included in the new `config-next.toml`
+7. the user, immediately after installing the new software, performs the config migration.
+8. when the current upgrade activation point is reached (see below for further details on how the running node will get
+   to know of the activation point) and the node has finished all its work in that era, it exits normally with an exit
+   code 0
+9. the `upgrade-bootstrapper` renames `casper-node-next` to `casper-node` (replaces the old binary)
+10. repeat from step 3
 
-Should the child process stop with a non-zero exit code at any stage, the `upgrade-bootstrapper` should itself panic.
-It is up to the user to restart the `upgrade-bootstrapper` manually.
+Should the child process stop with a non-zero exit code at any stage, the `upgrade-bootstrapper` should itself exit with
+an error.  It is up to the user to restart the `upgrade-bootstrapper` manually.
 
 This flow also allows for an installer or user to upgrade the `upgrade-bootstrapper` itself.  The `upgrade-bootstrapper`
 should be able to be killed, replaced and restarted while maintaining the integrity of the flow above.
@@ -61,17 +58,42 @@ location of the `casper-node` binary) we can handle these locations in a similar
 
 [reference-level-explanation]: #reference-level-explanation
 
+### Default locations
+
+The following default locations will be used:
+
+| File                               | Location                                      |
+| :--------------------------------- | :-------------------------------------------- |
+| `casper-node` binary               | `/var/lib/casper/bin/casper-node`             |
+| `upgrade-bootstrapper` binary      | `/var/lib/casper/bin/upgrade-bootstrapper`    |
+| `casper-node` config file          | `/etc/casper/config.toml`                     |
+| `upgrade-bootstrapper` config file | `/etc/casper/upgrade-bootstrappe-config.toml` |
+
 ### Changes to existing `casper-node` (running in validator mode)
 
-We require the following changes to `casper-node`:
+As well as the other changes specified in the sections below, we require the following changes to `casper-node`:
 
-* running in validator mode, when we reach the upgrade activation point, exit gracefully
+* running in validator mode, when we reach the upgrade activation point, exit gracefully.  Specifically, the point at
+  which the node should exit is one minute after it has received a third of finality signatures for the switch block.
+  The one minute delay is to allow the completion of gossiping messages for the benefit of other peers still awaiting
+  the requisite number of finality signatures.  It is an overly simple solution, pending replacement by a more robust
+  and efficient one
 * running in validator mode, if the node starts, joins and finds that it has already reached the upgrade activation
   point, exit gracefully
 * upgrade activation points should be defined in terms of era IDs, not block heights
 * config.toml updated to take the path to the chainspec upgrade file, similar to the `node.chainspec_config_path`
   option.  Alternatively, just remove this option, and require the chainspec, its upgrade file, accounts.csv and the
   config.toml file all to live in the same folder
+* chainspec loader will need to properly support returning an upgraded chainspec
+* initializer reactor will need to support calling the `commit_upgrade` method of contract runtime
+* consider creating an empty block immediately after committing an upgrade, in order to capture the global state hash
+  after possible data-migration but before the global state has been affected by executing any deploys.
+
+### Learning of a new upgrade activation point
+
+For simplicity's sake, the running `casper-node` will try on every era change to read from a file in a known location
+the next upgrade activation point.  This could be the chainspec upgrade file, or a standalone one containing just the
+activation point.
 
 ### Data migration
 
@@ -79,11 +101,14 @@ We require to add a new `migrate-data` subcommand to `casper-node`.  We could pr
 exclusively for data and config migration, but it seems simpler to just make this a subcommand.  Simpler for the
 installer, for the `upgrade-bootstrapper` and simpler to code.
 
+Currently we only require to migrate global state data.  The data held by the storage component can be left untouched
+as it won't be used after a major version protocol upgrade.
+
 Migration of any of the databases needs to take the following approach by iterating each value in the DB and doing:
 
 1. parse as the old type
 1. convert it to the new type
-1. store the new type, possibly under a different key
+1. store the new type, possibly under a different key, or in a new database
 1. if stored under a different key (i.e. it's not just an overwrite), then delete the old value
 
 We need to be able to resume migration in case it's killed while running.  This means we need to ensure that we either
@@ -91,37 +116,31 @@ keep track of our progress by writing to disk after each individual item is succ
 that parsing a new value as an old value fails, or else we ensure that running the conversion step on an
 already-converted item is a no-op.
 
-If data migration is going to take a prohibitively long time to execute, we should look to have the
-`upgrade-bootstrapper` perform _data_ migration only (not config migration) as soon as the installer provides the new
-`casper-node` binary, while the current `casper-node` is still running in validator mode.  However, this might result in
-a doubling or more of the disk space consumed by the databases, since the migrated data cannot be allowed to overwrite
-the existing data while the node is still running as a validator, and _this_ disk consumption might prove to be
-prohibitive.
-
-Data migration would also still need to be executed again once the old `casper-node` stops due to reaching
-the upgrade activation point, since more data will likely have been written to the databases between then and when the
-data migration was previously completed.  Running the `migrate-data` subcommand of the node could be done repeatedly at
-regular intervals to ensure that only a small amount is left to do after the old node binary has stopped.
+Data migration will only be executed once the old `casper-node` stops due to reaching the upgrade activation point, as
+there is no time saving to migrating the trie store prior to that point.
 
 ### Config migration
 
-Migration will also need to handle converting an old config file to a new one.  For this to work, we need to know which
-config values have been modified by the user from their default values.  This requires either default values to be
-provided, hard-coded in the node, or else the `migrate-data` subcommand will need to be passed not just the old and new
-config.toml files, but also a copy of each with their default values set.
+We require to add a new `migrate-config` subcommand to `casper-node`.  Config migration will handle converting an old
+config file to a new one.  For this to work, we need to know which config values have been modified by the user from
+their default values.  This requires either default values to be provided, hard-coded in the node, or else the
+`migrate-data` subcommand will need to be passed not just the old and new config.toml files, but also a copy of each
+with their default values set.
 
 We moved away from hard-coding defaults in the code a while ago, but a challenge to the other approach is ensuring the
 user _copies_ the installed vanilla config rather than renaming or deleting it.  Two possible solutions are:
 1. have the installer create a read-only copy of the default config exclusively for the `upgrade-bootstrapper`'s use,
-   and a separate writable one for the user to modify/copy/move as they wish
+   and a separate writeable one for the user to modify/copy/move as they wish
 1. require that every config value is annotated with a standard comment specifying the default value
+1. `include_str!` the default config.toml to make it available at compile time
 
-It seems like annotating the config file with default values might be somewhat more "normal" and should be fairly simple
-to implement and test.
+It seems like using `include_str!` might be somewhat more "normal" and should be fairly simple to implement and test.
 
 To migrate, the node will identify any modified values in the old config, and apply the same changes to the new config
 if the corresponding new config value has not already been modified by the user.
 
+It is expected that the user will execute this subcommand immediately after a new version is installed, and at least
+before it is required by the new `casper-node` binary.
 
 ### `upgrade-bootstrapper`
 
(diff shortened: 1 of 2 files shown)


text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,152 @@
...
+
+Under normal circumstances, the following steps will be performed:
+
+1. the `upgrade-bootstrapper` is started
+2. if the path to the node's `config.toml` was provided via an arg, the `upgrade-bootstrapper` writes the path value to
+   a hard-coded, known location (if writing fails, it panics).  Otherwise it reads the path value from the known
+   location.  If the path cannot be read from the known location, it panics
+3. the `upgrade-bootstrapper` spawns a child process, running the `casper-node` binary in validator mode, passing it the
+   path to its `config.toml` as an arg
+4. while they are running, before the network reaches the upgrade point, the next version of the `casper-node` gets
```
xcthulhu, line 29: What is an upgrade point?

text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,152 @@
...
+   installed, but not run.  The binary is always named `casper-node-next` and its config named `config-next.toml`.  Each
```
afck, line 30: Alternatively, we could call it `casper-node-<version>`, then multiple upgrades could be scheduled at the same time?

text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,152 @@
...
+   gets installed to the same location as the current equivalent files.  The location of the `config.toml` can be
+   retrieved from the known location used by the `upgrade-bootstrapper` at step 2.  An upgrade file for chainspec
+   modifications should also be installed and its path included in the new `config-next.toml`
+5. when the current upgrade activation point is reached and the node has finished all its work in that era, it exits
```
afck, line 34: So it must be possible to inform a running node about a new activation point — is that already implemented/ticketed?

text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,152 @@
...
+* running in validator mode, if the node starts, joins and finds that it has already reached the upgrade activation
+  point, exit gracefully
+* upgrade activation points should be defined in terms of era IDs, not block heights
+* config.toml updated to take the path to the chainspec upgrade file, similar to the `node.chainspec_config_path`
+  option.  Alternatively, just remove this option, and require the chainspec, its upgrade file, accounts.csv and the
+  config.toml file all to live in the same folder
+
+### Data migration
+
+We require to add a new `migrate-data` subcommand to `casper-node`.  We could provide a separate binary to be used
```
afck, line 72: An alternative would be to store the version number in the storage directory, so the new process knows what version the data belongs to and can determine whether it needs to perform the migration. (Not sure if that's better, but it would simplify the interface between the bootstrapper and the node.)

text/0032-upgrade-bootstrapper.md
Fraser999: It's an integer currently representing the block height at which a protocol upgrade should be applied.  I've heard that it would be more suitable to make this integer represent an era ID, since it would be more convenient to ensure protocol upgrades are only applied while transitioning between eras.

text/0032-upgrade-bootstrapper.md
Fraser999: I considered this, but wanted to try and keep the process as simple as possible for now.

The only use case I could think of for that was where a user has missed one or more intermediate updates.  So, e.g. their node was last running at version 1.0.0, but they've missed 2.0.0 and the network is currently on 3.0.0.

We could have the `upgrade-bootstrapper` search the install folder for the appropriate next package, but that adds a fair bit of complexity which I'm not sure is worthwhile.  In this case, the `upgrade-boostrapper` would probably only want to run the migration step of the v2.0.0 binary since there'd be no point running the node at v2.0.0 since it shouldn't be able to communicate with any of the peers at v3.0.0.  However, this might not be the case for a minor or patch level change.

If we think this overhead is indeed worthwhile, I can spend some further time updating the doc to try to cover these various flows.  However, I'd quite like to get general agreement on the current approach before spending time on this, in case it's wasted effort :)

text/0032-upgrade-bootstrapper.md
Fraser999: Well, probably, but not necessarily.  I suggested three approaches in the first [unresolved question](https://github.com/CasperLabs/ceps/blob/91457cd963dc87a495e5b73c2ba92d159d2c9039/text/0032-upgrade-bootstrapper.md#unresolved-questions), but there could likely be a better way than these.

text/0032-upgrade-bootstrapper.md
Fraser999: Yes, fair point.

My thinking was to try and keep the "normal" operation of node as far removed from the migration operation as possible, to the point where I considered a separate binary just for this.  Mainly because this should be the only place in our codebase where we need to handle two different versions of data or config types at the same time.  I'd like that sort of stuff to be kept away from the core of our current node codebase as practicable.  I was envisaging this not even touching our current storage component.

In terms of simplifying the interface, yes and no :)  The bootstrapper would only have to invoke the node once rather than twice.  However, to migrate the config data, it would need to take the path to the old config file too.  For the very first run, we'd either have to make the old path optional and not pass anything (slightly dangerous since it should always be required thereafter) or else we could just pass the same path twice, implying no migration is required (but that's a weird interface).

text/0032-upgrade-bootstrapper.md
afck: Ah, right, I was thinking of the third approach. I think my vote goes to that one. :+1:

afck (approved):

text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,152 @@
...
+exclusively for data and config migration, but it seems simpler to just make this a subcommand.  Simpler for the
+installer, for the `upgrade-bootstrapper` and simpler to code.
+
+Migration of any of the databases needs to take the following approach by iterating each value in the DB and doing:
+
+1. parse as the old type
+2. convert it to the new type
```
goral09, line 79: Since the new type is only part of the new binary, this has to be done using the `casper-node-new`. Will `casper-node-new` binary be able to access the LMDB files of the `casper-node` (old) while it's still running?

text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+### Changes to existing `casper-node` (running in validator mode)
+
+We require the following changes to `casper-node`:
+
+* running in validator mode, when we reach the upgrade activation point, exit gracefully
```
goral09, line 62: How does the _already-running_ node knows when it's time for an upgrade? Does it need to be restarted after node operator installs/updates the `upgrade-bootstraper`? Or does it monitor changes to some config file?

text/0032-upgrade-bootstrapper.md
Fraser999: I believe it should be safe to do that, although I describe that as a fallback option, only required if we decide data migration is going to take too long to be performed after the old binary stops.

If we take that approach, these steps will need adjusted so that we don't delete the old data while the old node binary is still running.

text/0032-upgrade-bootstrapper.md
Fraser999: Andreas had [the same question](https://github.com/CasperLabs/ceps/pull/32#discussion_r555655150) :)

goral09: So, the process will look as follows:
0. New upgrade activation point defined at Era=10,
1. Active validator finished Era=9, recognizes it's time to upgrade. Exists gracefully. 
2. `upgrade-bootstrapper` recognizes that `casper-node` exit so it will check if there's an upgrade scheduled. To do that, `casper-node` process has to signal this to the `upgrade-bootstraper` somehow – since `upgrade-bootstrapper` doesn't know what era is currently being active.
3. `upgrade-bootstraper` migrates the data. This can take a lot of time.
4. After migration is finished, `upgrade-bootstrapper` will perform the "replacement" of `casper-node` binaries.
5. `upgrade-bootstrapper` starts the new, upgraded binary.
6. `casper-node` starts a new era. Possibly a few minutes after the era started since the migration process takes time.


Does this sound correct?

text/0032-upgrade-bootstrapper.md
goral09: There's no third approach though on that list.

text/0032-upgrade-bootstrapper.md
Fraser999: Pretty much.  Minor details are:
* At step 2, the fact that `casper-node` exited with exit code 0 indicates an upgrade is due.  The node has no reason to exit otherwise, and a panic wouldn't return an exit code of 0.
* The migration is performed by the new binary rather than the old, so your steps 3 and 4 should swap places.

text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,152 @@
...
+while the current `casper-node` is still running in validator mode.  However, this might result in a doubling or more of
+the disk space consumed by the databases, since the migrated data cannot be allowed to overwrite the existing data while
+the node is still running as a validator, and _this_ disk consumption might prove to be prohibitive.  Data migration
+would also still need to be executed again once the old `casper-node` stops due to reaching the upgrade activation
+point, since more data will likely have been written to the databases between then and when the data migration was
+previously completed.
+
+### Config migration
+
+Migration will also need to handle converting an old config file to a new one.  For this to work, we need to know which
```
goral09, line 100: I am not sure this is necessary. Users are not allowed to modify chainspec, only the node configuration.

text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,152 @@
...
+its about to perform to the hard-coded known location immediately before performing the step.  On starting, the
+`upgrade-bootstrapper` will try to read this value and jump to that step.  If the value/file is absent, it assumes it's
+a fresh start.
+
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+Users lose the ability to override config values via command line.
```
goral09, line 136: Users can override only node configuration and those shouldn't change between protocol upgrades.

text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,152 @@
...
+1. How to get the upgrade activation point agreed by all validators?
+    * Could include first activation point in chainspec, and all chainspec upgrade files hold the _next_ activation
+      point. (Drawback is that we need to know much further in advance when we want the next activation point to occur)
+    * Once new chainspec upgrade file is installed, `upgrade-bootstrapper` could notify the node about its path (via
+      IPC or TCP listener on loopback).  Nodes could then use consensus to agree the point (could take a long time to
+      get consensus as install times of next version of software could vary widely)
+    * On each era change, `casper-node` could read from a file in a known location the next upgrade activation point.
+      This could be the chainspec upgrade file, or a standalone one containing just the activation point.
+
+1. Do we have to keep all chainspec upgrade files, or just the latest?  If just the latest, change the chainspec to not
```
goral09, line 152: I think we need to. If consensus needs `UNBONDING_DELAY` eras worth of information to properly validate consensus messages, and there will be upgrades within last `UNBONDING_DELAY` eras, each era may have different protocol configuration that defines whether a messages is valid or not. cc @afck

afck: I guess so… except if we say that cross-era slashing doesn't work across upgrade boundaries; maybe that's the easier option?

text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,153 @@
...
+[unresolved-questions]: #unresolved-questions
+
+1. How to get the upgrade activation point agreed by all validators?
+    * Could include first activation point in chainspec, and all chainspec upgrade files hold the _next_ activation
+      point. (Drawback is that we need to know much further in advance when we want the next activation point to occur)
+    * Once new chainspec upgrade file is installed, `upgrade-bootstrapper` could notify the node about its path (via
+      IPC or TCP listener on loopback).  Nodes could then use consensus to agree the point (could take a long time to
+      get consensus as install times of next version of software could vary widely)
+    * On each era change, `casper-node` could read from a file in a known location the next upgrade activation point.
+      This could be the chainspec upgrade file, or a standalone one containing just the activation point.
```
xcthulhu, line 150: @Fraser999 As you and I have discussed, I propose having this kept in global state rather than a file. While more complicated, this is guaranteed to be replicated across all of the nodes.

CC @henrytill WDYT?

text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,153 @@
...
+Migration will also need to handle converting an old config file to a new one.  For this to work, we need to know which
+config values have been modified by the user from their default values.  This requires either default values to be
+provided, hard-coded in the node, or else the `migrate-data` subcommand will need to be passed not just the old and new
+config.toml files, but also a copy of each with their default values set.
+
+We moved away from hard-coding defaults in the code a while ago, but a challenge to the other approach is ensuring the
+user _copies_ the installed vanilla config rather than renaming or deleting it.  Two possible solutions are:
+1. have the installer create a read-only copy of the default config exclusively for the `upgrade-bootstrapper`'s use,
+   and a separate writable one for the user to modify/copy/move as they wish
+1. require that every config value is annotated with a standard comment specifying the default value
```
xcthulhu, line 109:
```suggestion
2. require that every config value is annotated with a standard comment specifying the default value
```

text/0032-upgrade-bootstrapper.md
Fraser999: It's the node configuration I'm meaning here, not the chainspec.

xcthulhu:
Here are a few different scenarios I can think of and a plan for operating procedures in these cases.

1. In the event of a change to the wire format or networking layer, joining nodes will need a recent block hash from after the upgrade point.
2. In the event that the wire format and networking layer have not changed, the plan is to have _block headers_ never change format.  The block header will also contain a `protocol_version`.  The procedure for joining using a hash from before an upgrade is as follows:
   a) Download previous blocks from the block header in order to get the validators
   b) Get subsequent blocks, checking finality signatures from the validators along the way
   c) Stop getting blocks and copy state over when reaching a block header where `block_header.protocol_version == chainspec.protocol_version`

text/0032-upgrade-bootstrapper.md
goral09: cannot we copy over the old config and reuse it? if there are any new config parameters we can add them.

text/0032-upgrade-bootstrapper.md
Fraser999: I think that should be possible, and maybe even necessary occasionally.  Say we add a new field to the config file, like a path to a second secret key - they'd have to provide that info.

xcthulhu:
How are binaries being distributed?  I have suggested using IPFS.

goral09:
> How are binaries being distributed? I have suggested using IPFS.

I don't see why binaries cannot be distributed like they're now – either prebuilt on GitHub or let users build it themselves.

text/0032-upgrade-bootstrapper.md
henrytill: I think that's a bad idea.

text/0032-upgrade-bootstrapper.md
goral09: I didn't say we should not allow for overriding of node configuration but I see now what you meant by the original comment on "drawback" :+1:

text/0032-upgrade-bootstrapper.md
goral09: It seems so but is this a correct thing to do?

afck: Not quite, no. It does create a short window for unslashed equivocations. :confused:

goral09: It's not only about equivocations though – every running node should gossip (and validate) consensus messages from the past eras. If we do not store old chainspecs we basically say that after protocol upgrade we do not accept any old message. We're "cutting off" all the past eras effectively.

afck:
To clarify: This proposal would mean that usually during an upgrade all nodes go offline simultaneously, to restart, and possibly to do data migration?

text/0032-upgrade-bootstrapper.md
Fraser999: This isn't really required as the list is correctly displayed when rendered.  I tend to favour always using `1.` for each point as I don't have to renumber if I insert a point, although I favour explicit numbering if I'm explicitly referring to one point from another.  (I know I broke my own guidelines in the Data migration" section!)

Fraser999:
>> How are binaries being distributed? I have suggested using IPFS.
>
> I don't see why binaries cannot be distributed like they're now – either prebuilt on GitHub or let users build it themselves.

It's outside the scope of this CEP I believe, although ultimately I hoped we might be able to deliver new packages via our own P2P network.

Fraser999:
> To clarify: This proposal would mean that usually during an upgrade all nodes go offline simultaneously, to restart, and possibly to do data migration?

Yes, exactly.

text/0032-upgrade-bootstrapper.md
Fraser999: So, this proposal does really expect that nodes will not be able to communicate across protocol upgrades (at least major version change ones).  That's the main rationale for the proposal - to avoid the need for complex branching code which supports communications and data types which are incompatible.

I'd anticipated that a node would be able to communicate about previous eras using migrated data and using the new version of the message types.  For example, if we change the block structure in a version upgrade, migration would change old blocks to the new format, allowing them to still be gossiped.

Would that suffice for consensus, or do we _really_ need to support handling old versions of data and/or messages after a major protocol upgrade?

text/0032-upgrade-bootstrapper.md
Fraser999: Yes, but an issue is that we'd be hard-coding default values into the Rust code rather than providing defaults via config files as we currently do.  I think it's better to stick to how we do it currently to avoid changing too much.

goral09:
> > > How are binaries being distributed? I have suggested using IPFS.
> > 
> > 
> > I don't see why binaries cannot be distributed like they're now – either prebuilt on GitHub or let users build it themselves.
> 
> It's outside the scope of this CEP I believe, although ultimately I hoped we might be able to deliver new packages via our own P2P network.

How does it increase the security of the binaries?

goral09:
> > To clarify: This proposal would mean that usually during an upgrade all nodes go offline simultaneously, to restart, and possibly to do data migration?
> 
> Yes, exactly.

Do we expect and accept the downtime of the whole network for the time of an upgrade?

text/0032-upgrade-bootstrapper.md
goral09: It depends if we are OK sacrificing the some security of the protocol. Like @afck , in this case we open up ourselves to an attack where equivocators can continue unslashed. Personally, I think it's a bad idea and we should not do it.

goral09: > So, this proposal does really expect that nodes will not be able to communicate across protocol upgrades (at least major version change ones). 

Sounds more like a solution for an emergency restart of the network and not a protocol upgrade.

Fraser999:
> Do we expect and accept the downtime of the whole network for the time of an upgrade?

Well, I hedged my bets here, by stating the following in the [data-migration section](https://github.com/CasperLabs/ceps/blob/9a3f2b90a4eed025078f77b0564520b645feaeea/text/0032-upgrade-bootstrapper.md#data-migration):

> If data migration is going to take a prohibitively long time to execute, we should look to have the `upgrade-bootstrapper` perform _data_ migration only (not config migration) as soon as the installer provides the new `casper-node` binary, while the current `casper-node` is still running in validator mode.  However, this might result in a doubling or more of the disk space consumed by the databases, since the migrated data cannot be allowed to overwrite the existing data while the node is still running as a validator, and _this_ disk consumption might prove to be prohibitive.

So, basically, I was hoping that we could accept the downtime for simplicity's sake.  But if not, we'd need to take the alternative approach suggested.  Either the `upgrade-bootstrapper` will need to be able to handle that from the get go, or we'd need to upgrade the `upgrade-bootstrapper` itself first.

Your question pushed me to think in more detail about this and also the upgrading of the `upgrade-bootstrapper` itself.  I added a little more on both these things, including adding another unresolved question.  I'm now leaning more towards just having the more complex flow in place from the start.  I expect we're going to need it eventually, so probably best to try and design it now.

text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,153 @@
...
+
+[reference-level-explanation]: #reference-level-explanation
+
+### Changes to existing `casper-node` (running in validator mode)
+
+We require the following changes to `casper-node`:
+
+* running in validator mode, when we reach the upgrade activation point, exit gracefully
+* running in validator mode, if the node starts, joins and finds that it has already reached the upgrade activation
+  point, exit gracefully
```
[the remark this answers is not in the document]
Fraser999: > Can a node/validator be late to obtain the updated code?

As long as we're relying on humans to actively download and install updated software, I'd say yes.  Probably even if we do all we can to automate this in the future, we should still plan to handle this case.

> If so, it is possible for them to 'miss' the upgrade activation point? What happens then?

Probably the consensus team will be better placed to answer whether we can handle that gracefully.  I was anticipating it would be like a validator joining/re-joining after the genesis timestamp has expired.

If they've missed multiple upgrades, ultimately the brute force approach of treating that case like a brand new user just starting for the first time with our updated software should suffice.

> Additionally, for new nodes that join the network after the first or second upgrade has been processed, what does the sync and joining process look like? Can we lay this out?

@xcthulhu and I spoke about this, and we believe that if we ensure the block header is such that it will always hold the protocol version and the list of validators for the given era (at least in the switch blocks) then we should be able to make use of the current design for fast sync.  It might be worth having a meeting with interested parties to ensure we're all on the same page though.

Fraser999:
> Does it make sense for the user to be more hands on for the config migration? Meaning, should we consider a separate command to 'migrate the config-next' and have the user say 'yes it's ready' - I feel like migrating a text configuration file is rife with string parsing problems, etc... It will be hard for us to provide guarantees here. Perhaps if we are adding values, no issue, but if we need to change anything, it could get messy.

I think string parsing should be reasonably uncomplicated.  We're parsing the config files from toml format directly into native Rust structs using the serde crate.  It's pretty robust.

However, you made me think a bit further about the config migration process, and there probably _is_ room for improvement.  I can see some difficult edge cases cropping up.

Do you think it'd be reasonable to have the install process be the one which performs the config migration?

The user is already involved at that stage, and by doing it then, the upgrade-bootstrapper would not need to be paused or stalled awaiting a signal from the user that the new software is ready to be used.

So, the installer would invoke the new casper-node binary with a `migrate-config` subcommand (separate to the `migrate-data` subcommand suggested in the CEP) as a post-install step, passing it the old and new config paths, and on completion, ask the user to examine the new config for any changes they want to make manually.  If the old config doesn't exist, the post-install step would be skipped.

text/0032-upgrade-bootstrapper.md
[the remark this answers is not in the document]
goral09: It's not even required that the header holds the validator set as currently it doesn't and we're still able to sync – we read this information from the global state. Joining node will need a trusted hash that's after the last upgrade, that's all.

text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,171 @@
...
+      get consensus as install times of next version of software could vary widely)
+    * On each era change, `casper-node` could read from a file in a known location the next upgrade activation point.
+      This could be the chainspec upgrade file, or a standalone one containing just the activation point
+
+1. Do we have to keep all chainspec upgrade files, or just the latest?  If just the latest, change the chainspec to not
+   hold them in a `Vec`
+
+1. Should we design the `upgrade-bootstrapper` to run migration from the moment it detects that a new version  of
+   `casper-node` has been installed?  This minimises downtime while migration is performed after the old node binary
+   exits and before the new binary starts.
```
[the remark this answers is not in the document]
goral09: It complicates the migration process. First, it will migrate all the data that exists at the moment of installation, and later it will have to migrate the remaining data, that was added in-between the install and upgrade.

text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,171 @@
...
+the upgrade activation point, since more data will likely have been written to the databases between then and when the
+data migration was previously completed.  Running the `migrate-data` subcommand of the node could be done repeatedly at
+regular intervals to ensure that only a small amount is left to do after the old node binary has stopped.
+
+### Config migration
+
+Migration will also need to handle converting an old config file to a new one.  For this to work, we need to know which
+config values have been modified by the user from their default values.  This requires either default values to be
+provided, hard-coded in the node, or else the `migrate-data` subcommand will need to be passed not just the old and new
+config.toml files, but also a copy of each with their default values set.
```
fizyk20, line 111: How about we `include_str!` the default `config.toml` somewhere in the code? This way we aren't hardcoding the defaults per se, but the binary still has the defaults contained within it.

text/0032-upgrade-bootstrapper.md
```diff
@@ -0,0 +1,171 @@
...
+[unresolved-questions]: #unresolved-questions
+
+1. How to get the upgrade activation point agreed by all validators?
+    * Could include first activation point in chainspec, and all chainspec upgrade files hold the _next_ activation
+      point. (Drawback is that we need to know much further in advance when we want the next activation point to occur)
+    * Once new chainspec upgrade file is installed, `upgrade-bootstrapper` could notify the node about its path (via
+      IPC or TCP listener on loopback).  Nodes could then use consensus to agree the point (could take a long time to
+      get consensus as install times of next version of software could vary widely)
+    * On each era change, `casper-node` could read from a file in a known location the next upgrade activation point.
+      This could be the chainspec upgrade file, or a standalone one containing just the activation point
```
fizyk20, line 158: FWIW, I like this option best. Only I think the upgrade points were supposed to be included in the chainspec, so maybe we should keep them there, and re-read the chainspec at every era change?

darthsiroftardis (approved):

text/0032-upgrade-bootstrapper.md
Fraser999: Interesting idea - that could well be simpler.  The only downside I can see is that I quite liked the idea of improved UX by enforcing every config value to have user-facing documentation specifying the default value.  But that's a minor point - we can still include them where they're deemed useful.

text/0032-upgrade-bootstrapper.md
Fraser999: Well, we'd need to ensure we don't treat _that_ chainspec as current.  It seems like we could reuse the code which reads upgrade points from a file here rather than actually applying the upgrade point to the chainspec.

text/0032-upgrade-bootstrapper.md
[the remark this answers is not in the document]
Fraser999: We've decided against migrating data while the old node is running since there would be no reduction in downtime between versions as migrating the trie store doesn't benefit from a piecemeal approach like this.

xcthulhu:
@afck I chatted with @Fraser999 and we identified a lose end: we need to know what info needs to stay in the header to support the era supervisor.  This won't be addressed in this CEP, however.

afck:
> we need to know what info needs to stay in the header to support the era supervisor

I think it's just the random seed and the validator set?

dwerner (commented):
Read through, I like where this is going.

dwerner (approved):

xcthulhu (approved):

PR #33, opened 2021-01-13
Node Crash Recovery

fizyk20:
[Rendered](https://github.com/fizyk20/ceps/blob/crash-recovery/text/0033-crash-recovery.md)

### DIFF
--- a/text/0033-crash-recovery.md
+++ b/text/0033-crash-recovery.md
@@ -42,7 +42,7 @@ The consensus component should:
 
 - Whenever a unit or an endorsement is created, emit a request to storage to store the latest data for the era. The data stored would contain:
     - The hash of the latest unit
-    - The list of hashes of endorsed units
+    - The list of hashes of latest units by each validator endorsed by us
 
 This data should be keyed by the consensus instance ID.
 
@@ -81,9 +81,9 @@ Point 2 makes sure that at least one honest validator has received our unit befo
 
 Point 5 makes sure that we don't try starting from an older unit unless no validator seems able to send it to us.
 
-However, this scheme still has a weakness. It is theoretically possible that we will remove all the older hashes from the list when just a single honest validator confirms its reception, and that this validator will crash right afterwards. In such a case we will still be stuck with just hashes of units that no active honest validator has heard of and that we won't be able to sync. The risk should be significantly lower, though.
+However, this scheme still has an important weakness. If only some of the honest validators knew our unit before they all crashed, and the only ones that still have it are malicious, they could withhold this information, prompting us to start at some earlier point in the history and create a conflicting unit. Then they could reveal our old unit and claim that we equivocated.
 
-The question here is whether the complexity of this alternative is worth the decrease in the risk we would get by using it.
+Because of this risk, we probably shouldn't use this approach and go for the simpler, more secure one.
 
 ### Other alternatives
 
--- a/text/0033-crash-recovery.md
+++ b/text/0033-crash-recovery.md
@@ -4,7 +4,7 @@
 
 [summary]: #summary
 
-CEP PR: [casperlabs/ceps#0000](https://github.com/casperlabs/ceps/pull/0000)
+CEP PR: [casperlabs/ceps#0033](https://github.com/casperlabs/ceps/pull/33)
 
 This is an outline of a plan to allow nodes to rejoin eras that are already in progress, without risking to equivocate.
 
--- a/text/0033-crash-recovery.md
+++ b/text/0033-crash-recovery.md
@@ -63,6 +63,8 @@ Any sending of a unit / an endorsement should only happen after the storage requ
 
 Every creation of a unit or an endorsement will now involve storage - however, this seems unavoidable in that we need to store the knowledge of our activity in consensus in some permanent way.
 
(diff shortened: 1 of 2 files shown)


text/0033-crash-recovery.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+The consensus component should:
+
+- Whenever a unit or an endorsement is created, emit a request to storage to store the latest data for the era. The data stored would contain:
+    - The hash of the latest unit
+    - The list of hashes of endorsed units
```
afck, line 45: It would even suffice to, for each validator, store only the hash of their _latest_ unit that we endorsed.

text/0033-crash-recovery.md
```diff
@@ -0,0 +1,108 @@
...
+
+This data should be keyed by the consensus instance ID.
+
+In order to facilitate removing data for old eras (ones that are older than `UNBONDING_PERIOD`), we should also store a map of era ID → instance ID - as instance IDs themselves carry no information about how old the era is. We can, however, identify the era IDs of old eras, and using this map, we could also find the entries to be removed.
```
afck, line 49: Maybe an easier approach would be to put an expiration date into the file itself.
That sounds dangerous because we only know an _earliest_ date of when the validators will be unbonded, but not a latest one.

I think the solution is to define some constant, say "2 weeks", and:
* When creating a file for an era, put the expiration date into it, which is the earliest day where the era could expire, plus 2 weeks.
* Periodically delete all files with expiration date in the past.
* If we actually happen to instantiate an era whose expiration date would already be in the past, don't participate in it at all. (I.e. always assume we _did_ have a file and deleted it.)

text/0033-crash-recovery.md
```diff
@@ -0,0 +1,108 @@
...
+3. Whenever we receive a unit citing U[n] from another validators whose weights add up to more than our FTT, we remove the hashes of U[n-1] and older units from the list.
+4. When we rejoin an era and try to sync the state, request the latest unit from the list from other validators.
+5. If no validator has this unit, remove it from the list and try with the next latest one.
+
+Point 2 makes sure that at least one honest validator has received our unit before we remove its ancestors from our list. This way, we should always have at least one hash that corresponds to a unit received by an honest validator.
+
+Point 5 makes sure that we don't try starting from an older unit unless no validator seems able to send it to us.
+
+However, this scheme still has a weakness. It is theoretically possible that we will remove all the older hashes from the list when just a single honest validator confirms its reception, and that this validator will crash right afterwards. In such a case we will still be stuck with just hashes of units that no active honest validator has heard of and that we won't be able to sync. The risk should be significantly lower, though.
```
afck, line 84: Another drawback is that the only node with our latest unit could be malicious. They could withhold it from us until we drop it from the list and create an equivocating unit. Then they could publish it and have proof of us equivocating.

fizyk20: Right. I convinced myself at some point that this couldn't happen, but now I think you're right and actually it could. So good point :+1:

text/0033-crash-recovery.md
fizyk20: I would prefer not to rely on specific timestamps where we can't be certain that they will make sense, especially when we have other options. This approach might use less state than what I proposed, but I think it might turn out to be more complex in the end.

afck (approved):

text/0033-crash-recovery.md
```diff
@@ -0,0 +1,108 @@
...
+- If the entry exists:
+    - Pass all the data to the `ActiveValidator` instance.
+    - Only create new units if the unit with the latest hash is known to the consensus state.
+    - Only create new endorsements if all the previous endorsements are known to the consensus state.
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+Every creation of a unit or an endorsement will now involve storage - however, this seems unavoidable in that we need to store the knowledge of our activity in consensus in some permanent way.
```
goral09, line 64: I think we need to mention that if a node crashes between persisting the new unit and gossiping it, then upon restart it won't be able to join that era.

text/0033-crash-recovery.md
```diff
@@ -0,0 +1,108 @@
...
+
+In order to prevent situations in which we get stuck with a unit hash that no other validator has heard of, we could use a solution like the following:
+
+1. Let's denote n-th unit created by us by U[n].
+2. Whenever we create a unit, we _append_ its hash to the list of hashes of the units we created.
+3. Whenever we receive a unit citing U[n] from another validators whose weights add up to more than our FTT, we remove the hashes of U[n-1] and older units from the list.
+4. When we rejoin an era and try to sync the state, request the latest unit from the list from other validators.
+5. If no validator has this unit, remove it from the list and try with the next latest one.
+
+Point 2 makes sure that at least one honest validator has received our unit before we remove its ancestors from our list. This way, we should always have at least one hash that corresponds to a unit received by an honest validator.
```
goral09, line 80: Isn't that point 3?

goral09 (approved):
Despite known drawbacks in both of the proposed approaches, this seems good enough. 

There is one scenario I can think of, with the proposed approach, that would stall the consensus entirely:

Assuming we implemented the simpler solution (store hash of the latest unit/endorsement and request it upon restart), imagine that enough validators crashed, without gossiping their latest units, so that after a restart network can't finalize the switch block. Those validators won't be able to sync that era's protocol state, so they will wait until the next one, and the rest of the network will wait for more stake to finalize the switch block.

PR #34, opened 2021-01-14
Validator key management

ledgerleap:
Safe Script for Managing Signing Keys Vs. Withdraw Keys for Validators


AlexanderLimonov:
Please review CEP 31 (https://github.com/CasperLabs/ceps/pull/31) to see if it might cover your use cases.

MParlikar (commented):
It  appears that this feature  is addressed in [CEP#31 ](github.com/CasperLabs/ceps/pul/31/files)

PR #35, opened 2021-01-25
Emergency Restarts

fizyk20:
CEP for a better approach than the one outlined in CEP #6.

[Rendered](https://github.com/fizyk20/ceps/blob/emergency-restarts/text/0035-emergency-restarts.md)

### DIFF
--- a/text/0035-emergency-restarts.md
+++ b/text/0035-emergency-restarts.md
@@ -49,35 +49,25 @@ On the consensus side, it is important for an era change to happen at the bounda
 
 ### The structure of an upgrade point
 
-Currently, an upgrade point supports defining some new config values. We need it to also be able to specify some transforms that should be applied to the global state. This could be realized in multiple ways:
-
-1. Specifying the modifications as direct key-value pairs to be written to the global state (potentially overwriting the values under existing keys).
-2. Specifying arbitrary WASM to be executed, which would result in the transforms to be applied.
-3. For some special use cases, specifying changes in other ways, eg. support for changes to the validator set to be specified using an `accounts.csv`-like file.
-
-In this CEP, we suggest a combination of options 2 and 3.
+Currently, an upgrade point supports defining some new config values. We need it to also be able to specify some transforms that should be applied to the global state. In order to achieve this goal, it will be made possible for the upgrade points to optionally specify a file containing key-value pairs to be written to the global state (potentially overwriting the values under existing keys).
 
 ## Reference-level explanation
 
 [reference-level-explanation]: #reference-level-explanation
 
-The main thing required is to extend the `components::chainspec_loader::chainspec::UpgradePoint` structure to make it possible to specify arbitrary modifications to the global state.
-
-The best way of doing that is left as an unresolved question. We suggest adding the following two fields:
+The main thing required is to extend the `components::chainspec_loader::chainspec::UpgradePoint` structure to make it possible to specify arbitrary modifications to the global state. This will be achieved by adding a single field:
 
 ```rust
 struct UpgradePoint {
     // ... other fields
-    /// the paths to arbitrary WASM to be executed, in the specified order
-    upgrade_wasm_paths: Vec<String>,
-    /// the path to an `accounts.csv`-like file specifying the new set of validators
-    new_accounts: String,
+    /// The path to the file containing key-value pairs to be written to the global state.
+    state_write_spec: Option<String>,
 }
 ```
 
-These would have to be translated into analogous fields in `execution_engine::core::engine_state::upgrade::UpgradeConfig` (but probably containing the data itself, ie. WASM binary bytes and the contents of `accounts.csv` instead of the paths).
+It would have to be translated into an analogous field in `execution_engine::core::engine_state::upgrade::UpgradeConfig` (but containing the key-value map itself, not the path).
 
-The execution engine would execute the specified WASM and update the accounts and the state of the auction contract as a part of `EngineState::commit_upgrade`.
+The execution engine would execute the specified writes as a part of `EngineState::commit_upgrade`.
 
 The data about activation points would have to be made available to the `BlockExecutor` and the `EraSupervisor`. The BlockExecutor would have to be aware of them in order to request the execution engine to commit the upgrades at the right moments. The EraSupervisor would need this data to properly determine switch blocks.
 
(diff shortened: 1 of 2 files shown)


text/0035-emergency-restarts.md
```diff
@@ -0,0 +1,116 @@
...
+
+The criteria for validity of a block at height `h` should also take this into account, i.e. a block at height `h` is only valid if its global state hash is the result of applying the above steps to the global state specified at height `h-1`.
+
+Note that this has some desired properties:
+- We can specify any change to the state at any height without breaking the continuity of the chain. The first block after the restart is just like any other regular block, which means that eg. chain synchronization after the restart will work just like before the restart.
+- If the last safe block before the restart was at height `h`, but there were also blocks `h+1...h+n` that were finalized, they will become invalid after an upgrade point is specified at height `h+1`. Since the upgrade point wasn't included in the chainspec before the restart, the global state of the old block `h+1` won't satisfy the validity condition after an upgrade is added and the network is restarted. This means that all the blocks finalized after `h` will be discarded after restarting with the new chainspec and the network will resume operation at height `h+1`.
```
afck, line 44: But we should also take into account the possibility where we want to continue from a block at height `h` that in the view of some of the validators wasn't even seen as finalized.

I guess this can be achieved just by using the trusted hash?

text/0035-emergency-restarts.md
```diff
@@ -0,0 +1,116 @@
...
+
+The main thing required is to extend the `components::chainspec_loader::chainspec::UpgradePoint` structure to make it possible to specify arbitrary modifications to the global state.
+
+The best way of doing that is left as an unresolved question. We suggest adding the following two fields:
+
+```rust
+struct UpgradePoint {
+    // ... other fields
+    /// the paths to arbitrary WASM to be executed, in the specified order
+    upgrade_wasm_paths: Vec<String>,
```
EdHastingsCasperLabs, line 72: we don't need this concept any more, post EE-1050.

text/0035-emergency-restarts.md
```diff
@@ -0,0 +1,116 @@
...
+
+### The structure of an upgrade point
+
+Currently, an upgrade point supports defining some new config values. We need it to also be able to specify some transforms that should be applied to the global state. This could be realized in multiple ways:
+
+1. Specifying the modifications as direct key-value pairs to be written to the global state (potentially overwriting the values under existing keys).
+2. Specifying arbitrary WASM to be executed, which would result in the transforms to be applied.
+3. For some special use cases, specifying changes in other ways, eg. support for changes to the validator set to be specified using an `accounts.csv`-like file.
+
+In this CEP, we suggest a combination of options 2 and 3.
```
EdHastingsCasperLabs, line 58: I disagree; option 1 is preferable IMO.

goral09: I disagree with you Ed, I think specifying an upgrade as a list of `Transforms` (that are internal detail and were supposed to be removed already) is inferior. It will be only understood by the dev team. By using a combination of 2 and 3 we're not tying it to internal details and allowing anyone (that knows Rust) to read the upgrade. Writing `Transforms` "by hand" will also be a difficult task, we have no tooling for that.

text/0035-emergency-restarts.md
fizyk20: > I guess this can be achieved just by using the trusted hash?

I think so, yes.

text/0035-emergency-restarts.md
fizyk20: But that's not about the WASM for system contracts - it's for some code that would perform the changes to the global state that we require. These could also be deploys we'd want to replay from blocks that would become orphaned after a restart.

text/0035-emergency-restarts.md
fizyk20: With option 1 it would be hard to create the set of transforms (we'd need to create an additional tool for that, which we wanted to avoid) and hard to audit by the validators (which are supposed to reach a social consensus on the chainspec, so they should probably be aware of what is contained within). That's why my preference is for a combination of 2 and 3.

afck:
Another point I'd mention here is that we won't instantiate any Highway eras before the emergency restart point, and we won't do cross-era slashing across that point. (But we kind of decided that that will apply to regular upgrades as well, so maybe it doesn't need to be in this document after all.)

text/0035-emergency-restarts.md
```diff
@@ -0,0 +1,116 @@
...
+The main thing required is to extend the `components::chainspec_loader::chainspec::UpgradePoint` structure to make it possible to specify arbitrary modifications to the global state.
+
+The best way of doing that is left as an unresolved question. We suggest adding the following two fields:
+
+```rust
+struct UpgradePoint {
+    // ... other fields
+    /// the paths to arbitrary WASM to be executed, in the specified order
+    upgrade_wasm_paths: Vec<String>,
+    /// the path to an `accounts.csv`-like file specifying the new set of validators
```
goral09, line 73: Only validators? So this won't overwrite people's balances? We will use balances as they were at block `h-1` plus the potential changes from the `upgrade_wasm_paths`? 

I think we should stick to the Genesis-like approach where `accounts.csv` contains both validators' stakes and people's balances. 

cc @henrytill any opinion on this?

text/0035-emergency-restarts.md
```diff
@@ -0,0 +1,116 @@
...
+
+[prior-art]: #prior-art
+
+None at the moment.
+
+## Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+- How should the arbitrary transforms to the global state be specified in the chainspec? Is the proposal from this CEP good enough?
```
goral09, line 110: By "the proposal" you mean points 2 and 3 from *The structure of an upgrade point* ?

goral09 (approved):
No real blockers from me but I would like to see the question of how we overwrite/supply people's balances resolved.

text/0035-emergency-restarts.md
EdHastingsCasperLabs: That doesn't make sense to me. If you want to change global state directly the map of key, value concept is much more direct. If you want to execute deploys, you would call existing exec & commit from the contract_runtime component rather than try to burden the upgrade_commit method with the machinery to do that for you.

text/0035-emergency-restarts.md
EdHastingsCasperLabs: 1) We are not planning to get rid of Transforms afaik, so I'm curious why you think we are.
2) It wouldn't be Transforms; a map of Key, StoredValue is sufficient.
3) We do have tooling to write keyed values to global state; we use it all the time...the TrackingCopy. we would just loop over the map -> tc.write(k,v) applies the necessary state changes, get effects, commit, return the new root hash.

fizyk20: Re point 3, I (and Mateusz probably too) mean a tool that would generate such a map of Key → StoredValue from a more high-level description, like "reset the list of validators to this". I'm not sure how easy it would be to just manually write a list of keys to be modified in order to perform consistent changes like that, but I imagine it might be a bit complex and it would be easier to do that internally based on a description that is more human-readable.

text/0035-emergency-restarts.md
fizyk20: > If you want to change global state directly the map of key, value concept is much more direct & wasmless.

No disagreement here, my concern is how easy it would be to:
1. Prepare such an upgrade, so that it does what we want it to do (eg. reset the set of validators to some agreed upon one),
2. Audit such an upgrade. The validators would probably want to read the set of changes that will be applied and understand what effect that will have on the system. This can be hard to achieve with just a list of the keys and values that will be changed.

> If you want to execute deploys, you would extend the contract_runtime component to compose the EE's existing exec & commit methods using the actual deploys from the deploy store (nearly or perhaps exactly the same as we all ready do for normal execution) rather than try to burden the upgrade_commit method with the machinery to do that for you and without taking on extra complexity in supporting some special cased wasm.

If I understand you correctly, you mean that if we want to use WASM, we shouldn't be passing the deploys to `commit_upgrade`, but rather execute them and apply their results in some other place in the code, like the contract_runtime component? If yes, that sounds reasonable to me.

afck (approved):
Also no blockers from my side, but I agree with @goral09 (and I'm fine with either decision).

EdHastingsCasperLabs (approved):

PR #36, opened 2021-02-18
Reducing consensus memory usage.

afck:
[Rendered](https://github.com/afck/ceps/blob/consensus_mem/text/0036-consensus-memory-usage.md)

### DIFF
--- a/text/0036-consensus-memory-usage.md
+++ b/text/0036-consensus-memory-usage.md
@@ -5,7 +5,7 @@
 
 [summary]: #summary
 
-CEP PR: [casperlabs/ceps#0000](https://github.com/casperlabs/ceps/pull/0000)
+CEP PR: [casperlabs/ceps#0036](https://github.com/casperlabs/ceps/pull/0036)
 
 Drop past eras' protocol states earlier, to use less memory.
 
--- a/text/0000-consensus-memory-usage.md
+++ b/text/0000-consensus-memory-usage.md
@@ -1,0 +1,130 @@
+# Reducing consensus memory usage
+
+
+## Summary
+
+[summary]: #summary
+
+CEP PR: [casperlabs/ceps#0000](https://github.com/casperlabs/ceps/pull/0000)
+
+Drop past eras' protocol states earlier, to use less memory.
+
+
+## Motivation
+
+[motivation]: #motivation
+
+The network's validators are determined by an auction: If you want to become a validator, you send money — your bid — to the auction contract. You can add more money and increase your bid at any time. Withdrawing money reduces your bid immediately, but the payout happens with a delay of `U` eras, the _unbonding delay_. At the end of each era `E`, the highest bidders are selected as the validators for era `E + A + 1`, with a weight proportional to their bids. `A` is the _auction delay_.
+
+So in era `E`, the validators' weights are proportional to staked tokens that will be paid out no earlier than at the end of era `E + U - A`. Thus `B = U - A` is the number of past eras whose validators are still bonded and can still be slashed, and whose signatures therefore can still be trusted.
+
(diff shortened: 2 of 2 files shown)


xcthulhu:
Having chatted with @afck regarding this, I think swapping whole era states to disk is complicated and would not be necessary given the proposed changes.

Moreover, while it could be an optimization to back protocol state in a database, I don't see how we could quickly query partial order that arise from the DAG without reimplementing something like `ltree` indices from postgres.  This is a lot of added scope when the proposal here may be sufficient.

Fraser999 (approved):

text/0036-consensus-memory-usage.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+We drop an era's protocol state from memory as soon as all its blocks have a sufficient amount of finality signatures, and consider those signatures a replacement for the protocol state: just like the state, they are a proof of finality.
+
+Conflicting signatures cause the signer to get slashed, so that — just like with equivocations in the Highway protocol state — the finality signatures are actually backed by the stake.
+
+If an attacker creates an equivocation in an old era's Highway protocol, either it will have no effect, because everyone has already received the signed blocks and dropped the protocol state; or someone who still has the protocol state will detect the equivocation and still cause the attacker to get slashed.
```
goral09, line 40: Does it mean we need to detect equivocation between finality signature and the conflicting unit (consensus message)?

text/0036-consensus-memory-usage.md
```diff
@@ -0,0 +1,130 @@
...
+[prior-art]: #prior-art
+
+TBD
+
+
+## Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+None so far.
```
goral09, line 123: Shouldn't we still answer the question how to incentivize validators to create signatures?

goral09 (commented):
Good write, nice solution. I like this.

text/0036-consensus-memory-usage.md
afck: I was wondering about that as well and [on Wednesday we kind of agreed that we don't](https://casperlabs.atlassian.net/wiki/spaces/EN/pages/250282071/Consensus+design+meetings#2021-02-24).

Basically: If the validators reach consensus anyway, without using Highway for it, then we can't really complain — they _have_ reached consensus and signed a consistent blockchain, after all. :man_shrugging: 
(Anyway, not sure about this — let me know your thoughts!)

text/0036-consensus-memory-usage.md
afck: I'd consider it beyond the scope of this CEP, which is about reducing memory usage. Incentivizing signatures is something we need to implement anyway, independent of this CEP. A proposal was also made [in Wednesday's meeting](https://casperlabs.atlassian.net/wiki/spaces/EN/pages/250282071/Consensus+design+meetings#2021-02-24):

* There is no _additional_ reward for finality signatures; instead we make the regular block reward conditional on them.
* Proposed blocks (“proto blocks”) can depend on finality signatures.
* Instead of at the end of era 5, you receive your block rewards for era 5 only in the first switch block that depends (directly or indirectly) on your finality signature on the switch block of era 5 (which in turn depends on the signatures on all its ancestors).

text/0036-consensus-memory-usage.md
afck: Although, now that I think of it: It means that we lose the security guarantee for clients following the Highway protocol at a custom fault tolerance threshold. Seeing a block finalized with 10% does _not_ mean anymore that the validators would have to forfeit 10% of the stake to finalize a conflicting block, if they can just ignore Highway and create finality signatures for something else.

Let's discuss next Wednesday.

PR #37, opened 2021-03-08
Network Upgrades

Fraser999:
[Rendered](https://github.com/Fraser999/ceps/blob/NDRS-952-upgrades/text/0037-node-upgrades.md).


goral09:
Who is this document for? Users (validators) or developers? If the former, then I would say there are too many implementation details in it (like _set state to `RunNodeAsValidator`_), could we change the wording so that it makes sense to validators and not only devs?

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,95 @@
...
+they become active.
+
+The node has two relevant modes of operation: "validator mode", and "migrate-data mode".  The normal mode is validator.
+The migrate-data mode is used immediately after an upgrade to migrate any data stored on disk as required.
+
+The node also has three possible exit codes: 0, 101 or 102.  0 indicates success, 101 indicates an error (e.g. a panic)
+and 102 indicates the node should be downgraded to a previous version.  These exit codes are read and reacted to by the
+launcher.
+
+On startup, the launcher either tries to read its previously cached state from disk, or assumes a fresh start.  On a
```
goral09, line 27: Can you include information about what kind of state it caches?

Fraser999: Done in [b70fb7f](https://github.com/CasperLabs/ceps/pull/37/commits/b70fb7fee444b48e66112fda0a710b8c83a3ca85).

Fraser999:
> Who is this document for? Users (validators) or developers?

It's for developers primarily, since the reference-level explanation should give enough details to implement it.

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+* AP = activation point
+* current X = value taken from current chainspec
+* next X = value taken from next upgraded chainspec
+
+| Condition                                           | Scenario                                     | Action                                   |
+| :-------------------------------------------------- | :------------------------------------------- | :--------------------------------------- |
+| no blocks in storage, and PV 1.0.0                  | initial run                                  | commit genesis and continue running      |
+| no blocks in storage, and PV not 1.0.0              | invalid initial run                          | exit with code 102 to downgrade          |
+| HB.era + 1 == current AP and HP is switch block     | restart immediately after upgrade            | commit upgrade and continue running      |
+| HB.era + 1 == current AP and HP is not switch block | invalid run - blocks missing from storage    | exit with code 102 to downgrade          |
```
afck, line 72:
```suggestion
| HB.era + 1 == current AP and HB is switch block     | restart immediately after upgrade            | commit upgrade and continue running      |
| HB.era + 1 == current AP and HB is not switch block | invalid run - blocks missing from storage    | exit with code 102 to downgrade          |
```

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+| HB.era < current AP                                 | invalid run - blocks missing from storage    | exit with code 102 to downgrade          |
```
afck, line 73:
```suggestion
| HB.era + 1 < current AP                             | invalid run - blocks missing from storage    | exit with code 102 to downgrade          |
```
(I guess that's equivalent because of the order, but it might still be clearer.)

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+| HB.PV != current PV                                 | restart after fork                           | treat as hard-reset and continue running |
+| HB.PV == current PV and no next chainspec available | unplanned shutdown with no scheduled upgrade | continue running                         |
+| HB.era < next AP                                    | unplanned shutdown before scheduled upgrade  | continue running                         |
```
afck, line 76:
```suggestion
| HB.era + 1 < next AP                                | unplanned shutdown before scheduled upgrade  | continue running                         |
```

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+
+| Condition                                           | Scenario                                     | Action                                   |
+| :-------------------------------------------------- | :------------------------------------------- | :--------------------------------------- |
+| no blocks in storage, and PV 1.0.0                  | initial run                                  | commit genesis and continue running      |
+| no blocks in storage, and PV not 1.0.0              | invalid initial run                          | exit with code 102 to downgrade          |
+| HB.era + 1 == current AP and HP is switch block     | restart immediately after upgrade            | commit upgrade and continue running      |
+| HB.era + 1 == current AP and HP is not switch block | invalid run - blocks missing from storage    | exit with code 102 to downgrade          |
+| HB.era < current AP                                 | invalid run - blocks missing from storage    | exit with code 102 to downgrade          |
+| HB.PV != current PV                                 | restart after fork                           | treat as hard-reset and continue running |
```
afck, line 74: Shouldn't we restart and run the corresponding version?

text/0037-node-upgrades.md
afck, line 72: And should these be `next AP` instead?

afck (commented):
I'm still rather confused by the table. :dizzy_face: 
I added a few comments, but maybe I'm somehow misunderstanding it.

EdHastingsCasperLabs (approved):
Assuming typos are corrected...

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+
+The node also has three possible exit codes: 0, 101 or 102.  0 indicates success, 101 indicates an error (e.g. a panic)
+and 102 indicates the node should be downgraded to a previous version.  These exit codes are read and reacted to by the
+launcher.
+
+On startup, the launcher either tries to read its previously cached state from disk, or assumes a fresh start.
+Information cached includes which mode of operation the node was executed in, the protocol version, the path to the node
+binary and the path to the config folder.
+
+On a fresh start, the launcher searches for the lowest installed version of `casper-node` and starts running it in
```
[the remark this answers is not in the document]
Fraser999: That wouldn't be classed as a fresh start - there would be cached state and the launcher would resume operations using whatever node version was specified in the cached info.

A fresh start here implies a fresh install.

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+The `casper-node-launcher` is a relatively simple Rust binary which runs the `casper-node` binary as a child process.
+Normal operation will see the launcher running indefinitely, cycling upwards through versions of the node as and when
+they become active.
+
+The node has two relevant modes of operation: "validator mode", and "migrate-data mode".  The normal mode is validator.
+The migrate-data mode is used immediately after an upgrade to migrate any data stored on disk as required.
+
+The node also has three possible exit codes: 0, 101 or 102.  0 indicates success, 101 indicates an error (e.g. a panic)
```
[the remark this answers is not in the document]
Fraser999: To manually stop a casper-node, the user would call `systemctl stop` which causes the launcher to terminate the node.  If we want to change from that, it should probably be discussed in a separate CEP?

text/0037-node-upgrades.md
Fraser999: The scenario here is that the chain has forked somehow, so we delete the blocks back to the start of the era specified in the current chainspec and continue running.  If, since we last ran, the network has also upgraded, this should result in the sync process catching up and then shutting down for the launcher to then start the next node version.  However, if the network hasn't upgraded since then, there'd be no need - we're already running the correct version.

text/0037-node-upgrades.md
Fraser999: No, I don't think so.

When we schedule an upgrade for activation point era `x`, our node should run until the end of era `x - 1`.  This entry describes what happens when the next version of the node starts: its chainspec will have activation point `x` and its highest block in storage will have era `x - 1`.  On startup, there's a high chance that `next AP` will be `None` as we're unlikely to schedule and install two upgrades at the same time, outside of the scenario of a completely fresh install, where every previous version must be installed.

Fraser999: Fixed in [d62ffc1](https://github.com/CasperLabs/ceps/pull/37/commits/d62ffc1d930e7d2dc6d5d14659e13f144b5b4308).

text/0037-node-upgrades.md
Fraser999: Fixed in [d62ffc1](https://github.com/CasperLabs/ceps/pull/37/commits/d62ffc1d930e7d2dc6d5d14659e13f144b5b4308).

text/0037-node-upgrades.md
Fraser999: Fixed in [d62ffc1](https://github.com/CasperLabs/ceps/pull/37/commits/d62ffc1d930e7d2dc6d5d14659e13f144b5b4308).

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+
+* HB = highest block
+* PV = protocol version
+* AP = activation point
+* current X = value taken from current chainspec
+* next X = value taken from next upgraded chainspec
+
+| Condition                                           | Scenario                                     | Action                                   |
+| :-------------------------------------------------- | :------------------------------------------- | :--------------------------------------- |
+| no blocks in storage, and PV 1.0.0                  | initial run                                  | commit genesis and continue running      |
```
afck, line 69: In the future, we could add an optional "starting protocol version" to the chainspec, so you could e.g. launch a private network without having to start from 1.0.0.

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+| no blocks in storage, and PV not 1.0.0              | invalid initial run                          | exit with code 102 to downgrade          |
+| HB.era + 1 == current AP and HB is switch block     | restart immediately after upgrade            | commit upgrade and continue running      |
+| HB.era + 1 == current AP and HB is not switch block | invalid run - blocks missing from storage    | exit with code 102 to downgrade          |
+| HB.era + 1 < current AP                             | invalid run - blocks missing from storage    | exit with code 102 to downgrade          |
+| HB.PV != current PV                                 | restart after fork                           | treat as hard-reset and continue running |
+| HB.PV == current PV and no next chainspec available | unplanned shutdown with no scheduled upgrade | continue running                         |
+| HB.era + 1 < next AP                                | unplanned shutdown before scheduled upgrade  | continue running                         |
+| HB.PV != next PV                                    | missed an upgrade and forked                 | treat as hard-reset and continue running |
```
afck, line 77: If `HB.PV >= next PV`, would it be safe to also exit with code 0 and upgrade? Because e.g. we might be running version 1.1.0, even though the blocks on disk already contain all blocks of 1.2.0 and some of 1.3.0?

afck (approved):

text/0037-node-upgrades.md
Fraser999: Well, I can't really see a good reason for wanting that, but it's certainly do-able :)  Either way, it's probably outside the scope of this CEP, since it's not really related to network upgrades?

text/0037-node-upgrades.md
Fraser999: I think the safer option is to assume the blocks for 1.2.0 and 1.3.0 are not valid and should be deleted.  If they _were_ valid blocks, the user has just wasted some time and energy, nothing more.

I'm actually not sure how a user's setup could end up in that sort of a state without some jiggery-pokery.  They'd have to have e.g. tinkered with the launcher's cached state to make it start running node 1.1.0 rather than 1.3.0, or imported the LMDB files with the newer blocks from some other source.

text/0037-node-upgrades.md
xcthulhu, line 23:
```suggestion
The `casper-node` can emit three possible exit codes: 0, 101 or 102.  0 indicates success, 101 indicates an error (e.g. a panic)
```

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+The `casper-node-launcher` is a relatively simple Rust binary which runs the `casper-node` binary as a child process.
+Normal operation will see the launcher running indefinitely, cycling upwards through versions of the node as and when
+they become active.
+
+The node has two relevant modes of operation: "validator mode", and "migrate-data mode".  The normal mode is validator.
+The migrate-data mode is used immediately after an upgrade to migrate any data stored on disk as required.
+
+The node also has three possible exit codes: 0, 101 or 102.  0 indicates success, 101 indicates an error (e.g. a panic)
+and 102 indicates the node should be downgraded to a previous version.  These exit codes are read and reacted to by the
+launcher.
```
xcthulhu, line 25:
```suggestion
`casper-node-launcher`.
```

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+
+On startup, the launcher either tries to read its previously cached state from disk, or assumes a fresh start.
```
xcthulhu, line 27:
```suggestion
On startup, the `casper-node-launcher` either tries to read its previously cached state from disk, or assumes a fresh start.
```

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+validator mode.
+
+If the launcher cannot find an appropriate version at any stage of upgrading or downgrading, it exits with an error.
+
+
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+### casper-node-launcher
```
xcthulhu, line 41:
```suggestion
### `casper-node-launcher`
```

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+
+The following describes the control flow in the launcher:
+
+![control flow in the launcher](images/0037/casper-launcher-flow.png "control flow in the launcher")
+
+### casper-node
```
xcthulhu, line 47:
```suggestion
### `casper-node`
```

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+The node also has three possible exit codes: 0, 101 or 102.  0 indicates success, 101 indicates an error (e.g. a panic)
+and 102 indicates the node should be downgraded to a previous version.  These exit codes are read and reacted to by the
+launcher.
+
+On startup, the launcher either tries to read its previously cached state from disk, or assumes a fresh start.
+Information cached includes which mode of operation the node was executed in, the protocol version, the path to the node
+binary and the path to the config folder.
+
+On a fresh start, the launcher searches for the lowest installed version of `casper-node` and starts running it in
+validator mode.
```
xcthulhu, line 32: This behavior is a bit confusing for newcomers who want to run a node.

It really only makes sense at genesis.  Since genesis has already occurred, we should consider bundling a trusted hash with fresh installs.  Perhaps in the long run we could do something else if this is judged to be a non-theoretical security hazard.

In that case it would start in *joiner* mode.

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+The node checks for a newer node version having been installed on every new block created.  If it finds an upgrade, it
+reads the chainspec associated with the new version to establish the activation point.  It then schedules itself to exit
+at the end of the era immediately before that activation point.
+
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+1. We don't have the ability to slash across major protocol version upgrades.  This is not a major problem; only
```
xcthulhu, line 89:
```suggestion
  - We don't have the ability to slash across major protocol version upgrades.  This is not a major problem; only
```

text/0037-node-upgrades.md
```diff
@@ -0,0 +1,98 @@
...
+### casper-node-launcher
+
+The following describes the control flow in the launcher:
+
+![control flow in the launcher](images/0037/casper-launcher-flow.png "control flow in the launcher")
+
+### casper-node
+
+On startup, the node needs to decide what action to take depending upon whether it is an initial run (without fast sync
+this will always be treated as running from genesis), a restart after an upgrade, or a restart after an unexpected
```
xcthulhu, line 50: With fast sync this will still be treated as running from genesis if a trusted hash is not provided...

text/0037-node-upgrades.md
Fraser999: Bundling a trusted hash for convenience makes sense to me, but is outside of the scope of this CEP I think.

PR #38, opened 2021-03-19
NDRS-1019: Equivocator Policy

xcthulhu:
[Rendered](https://github.com/xcthulhu/ceps/blob/NDRS-1019/text/0038-equivocator-policy.md)


MParlikar (approved):
Looks good to me.  Future considerations should include forcibly unbonding an equivocating node & placing them in the unbonding wait period.

EdHastingsCasperLabs (approved):

PR #39, opened 2021-04-01
[smart-contracts] Dictionary

mpapierski:
[Rendered](https://github.com/mpapierski/ceps/blob/ee-1212-local-key-proposal/text/0039-dictionary.md)


darthsiroftardis (approved):

text/0039-local-keys.md
```diff
@@ -0,0 +1,153 @@
...
+[prior-art]: #prior-art
+
+In the past, we used to have `Key::Local,` which we first deprecated and then removed as we used it only in system contracts, and there was a better functionality in place to replace uses. This proposal builds on top of past implementation by fixing problems that it had along the way.
+
+## Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+- How do we name this feature to reflect better what it does? 
+- Does it make sense to make the provisioned URef a unit? Can we use it to store additional information related to the stored data under the local key?
```
goral09, line 145: How does the migration of local storage look like? How will the upgraded contract access the pre-upgraded local data?

mpapierski: This data migration will be transparent to already deployed contracts. I see this as an internal improvement, but the API would be forward-compatible.

During migration, we will iterate over reachable URefs (i.e., those stored under account or contract's named keys) detect if there's extra payload under that Unit with all the key bytes, then reindex the keys from the old hashed address (`hash(uref_addr || key_bytes)`), into `uref_addr || hash(key_bytes)` and remove the extra payload.

goral09: OK, thanks. I just wanted to make sure this is not being overlooked and we already have a plan for how to solve it. We can't deliver local keys without also allowing contract owners to transfer them between contract upgrades.

text/0039-local-keys.md
```diff
@@ -0,0 +1,153 @@
...
+fn list_local(uref) -> Vec<Vec<u8>>;
+```
+
+This function lists all keys under a Key::Local referenced by "URef." Internally it reads all the keys from the secondary index and returns them to the caller.
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+As explained above current limitation of key lengths inside a trie requires us to maintain a secondary index of keys under a URef to retain the ability to migrate into the unhashed, more extended keyspace for more efficient iteration.
```
xcthulhu, line 126: This restriction could be lifted.  @henrytill is it worth getting support for variable length keys so we can avoid maintaining this secondary index, or should that be done at a different time?

text/0039-dictionary.md
```diff
@@ -0,0 +1,118 @@
...
+```
+blake2b256(uref_addr || key_bytes)
+```
+
+Further operations such as reading or writing use the earlier algorithm to access the `Key::Dictionary` space.
+
+### dictionary_new
+
+```rust
+fn dictionary_new(name: &str) -> URef;
```
goral09, line 69: In the implementation it's `new_dictionary`.

text/0039-dictionary.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+Host provisions an URef of "unit" type under the hood used internally to secure access to the dictionary. Resulting URef will be automatically put into named keys, retaining his ability to add new data.
```
goral09, line 72: What happens if `name` is empty? Or if there already exists an entry with the same name in named keys collection of the account?

RitaMAllenCA:
@darthsiroftardis is this still relevant or can it be merged?  Please advise.....

darthsiroftardis (approved):

PR #41, opened 2021-04-14
Extensible Block Headers and Bodies

fizyk20:
[Rendered](https://github.com/fizyk20/ceps/blob/extensible-block-body/text/0041-extensible-block-body.md)

### DIFF
--- a/text/0041-extensible-block-body.md
+++ b/text/0041-extensible-block-body.md
@@ -85,6 +85,8 @@ An alternative allowing us to expand the block format would be to just perform a
 
 The CEP also proposes to keep the old blocks in the old format. An alternative would be to migrate the old database of the block bodies to the new format. However, since block body hashes are immutable, the keys after the migration wouldn't always be the hash of the value stored under that key, which would make checking the integrity of storage much harder, if not impossible.
 
+Another alternative would be not to store partial block bodies at all. For the purposes of preventing replay attacks, we only need to know which deploys were included in past blocks. This information could be stored in the form of a `deploy_hash → block_hash` database. We currently have such a map as an index (so this data is lost on node restart and requires the block body data to be repopulated), and a `deploy_metadata` database that stores the block hash along with execution results - the drawback of the latter is that it only stores entries for _executed_ deploys, and fast sync wouldn't execute past deploys, so this database wouldn't be populated.
+
 ## Prior art
 
 [prior-art]: #prior-art
--- a/text/0041-extensible-block-body.md
+++ b/text/0041-extensible-block-body.md
@@ -4,7 +4,7 @@
 
 [summary]: #summary
 
-CEP PR: [casperlabs/ceps#0000](https://github.com/casperlabs/ceps/pull/0000)
+CEP PR: [casperlabs/ceps#0041](https://github.com/casperlabs/ceps/pull/0041)
 
 Modify the internal structure of the block body, so that it will be easy to add new fields to it in the future in a backwards-compatible manner.
 
--- a/text/0041-extensible-block-body.md
+++ b/text/0041-extensible-block-body.md
@@ -12,9 +12,10 @@ This CEP proposes allowing for extensible block headers and bodies.  To achieve 
 
 [motivation]: #motivation
 
-There are two main reasons this will be useful:
-1. The block header and body will representable as content-addressed linked lists. It will be easy to add more fields to them should we ever need them.
+There are three main reasons this will be useful:
+1. The block header and body will be representable as content-addressed linked lists. It will be easy to add more fields to them should we ever need them.
 2. This is important for Fast Sync to be able to securely download only the relevant parts of old blocks, increasing performance.
+3. This will reduce the need for special handling of different versions of block headers and bodies.
 
 ## Guide-level explanation
 
@@ -245,8 +246,9 @@ impl BlockHeader {
         let hashed_era_end = match era_end {
             None => SENTINEL,
             Some(era_end) => era_end.hash(),
-        };        let hashed_height = hash(height.to_bytes());
-        let hashed_timestamp =hash(timestamp.to_bytes());
+        };
+        let hashed_height = hash(height.to_bytes());
+        let hashed_timestamp = hash(timestamp.to_bytes());
         let hashed_protocol_version = hash(protocol_version.to_bytes());
         let hashed_random_bit = hash(random_bit.to_bytes());
 
--- a/text/0041-extensible-block-body.md
+++ b/text/0041-extensible-block-body.md
@@ -52,8 +52,10 @@ This can be illustrated by the following diagram:
                          /                  \
                    hash(transfer_hashes)     hash_3_______
                                             /             \
-                                         hash(proposer)   hash("")
+                                         hash(proposer)   SENTINEL
 ```
+
+`SENTINEL` is a value that is very improbable to be a valid hash of some data - like, for example, the all-zeros hash (`000000...00`).
 
 So, for example, to securely download only the transfer hashes stored in a block with a known body hash, we will additionaly require only the hash of the deploy hashes list and `hash_3`. This will allow us to calculate `hash_2` and `body_hash`, and to verify that the body hash matches the one we have.
 
@@ -96,14 +98,14 @@ The second is a primitive for hashing slices. This follows the scheme for hashin
 pub fn hash_slice_rfold(slice: &[Digest]) -> Digest {
     slice
         .iter()
-        .rfold(hash(&[]), |prev, next| hash_pair(next, &prev))
+        .rfold(SENTINEL, |prev, next| hash_pair(next, &prev))
 }
 ```
 
 In this case `hash_slice_rfold(&[a, b, c])` effectively expands to:
 
 ```rust
-hash_pair(a, &hash_pair(b, &hash_pair(c, &hash(&[]))))
+hash_pair(a, &hash_pair(b, &hash_pair(c, &SENTINEL)))
 ```
 
 This scheme is suited to simple Merkle proofs and verification.  Hashing with a proof is almost identical to `hash_slice_rfold`.
@@ -121,7 +123,7 @@ The third hashing primitive constructs a [Merkle tree][2]. It is intended for ve
 ```rust
 fn hash_vec_merkle_tree(vec: Vec<Digest>) -> Digest {
     if vec.is_empty() {
-        return hash(&[]);
+        return SENTINEL;
     };
     let mut vec = vec;
     let mut k = vec.len();
@@ -178,7 +180,7 @@ where
     T: ToBytes,
 {
     match maybe_t {
-        None => Ok(hash(&[])),
+        None => Ok(SENTINEL),
         Some(t) => Ok(hash(t.to_bytes()?)),
     }
 }
@@ -241,7 +243,7 @@ impl BlockHeader {
 
         let hashed_era_id = hash(era_id.to_bytes());
         let hashed_era_end = match era_end {
-            None => hash(&[]),
+            None => SENTINEL,
             Some(era_end) => era_end.hash(),
         };        let hashed_height = hash(height.to_bytes());
         let hashed_timestamp =hash(timestamp.to_bytes());
--- a/text/0000-extensible-block-body.md
+++ b/text/0000-extensible-block-body.md
@@ -1,0 +1,104 @@
+# Extensible Block Body
+
+## Summary
+
+[summary]: #summary
+
+CEP PR: [casperlabs/ceps#0000](https://github.com/casperlabs/ceps/pull/0000)
+
+Modify the internal structure of the block body, so that it will be easy to add new fields to it in the future in a backwards-compatible manner.
+
+## Motivation
+
+[motivation]: #motivation
+
+There are two main reasons this will be useful:
+1. As the block body will essentially become a content-addressed linked list, it will get really easy to add more fields to it, should we ever need them.
+2. This is important for Fast Sync to be able to securely download only the relevant parts of old blocks, increasing performance.
+
+## Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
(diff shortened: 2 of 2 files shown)


afck (approved):

text/0041-extensible-block-body.md
```diff
@@ -0,0 +1,104 @@
+# Extensible Block Body
+
+## Summary
+
+[summary]: #summary
+
+CEP PR: [casperlabs/ceps#0041](https://github.com/casperlabs/ceps/pull/0041)
+
+Modify the internal structure of the block body, so that it will be easy to add new fields to it in the future in a backwards-compatible manner.
```
goral09, line 9: Is it really about modifying the internal structure of the block body or only about how its hash is computed?

text/0041-extensible-block-body.md
```diff
@@ -0,0 +1,104 @@
...
+- A `transfer_hashes` database would store the lists of transfer hashes keyed by `hash(transfer_hashes)`.
+- A `proposer` database would store the proposers keyed by `hash(proposer)`.
+
+The new way of hashing and new databases would start being used in a specific, chosen protocol version. The blocks in this protocol version and above would have their bodies hashed using the new way and be stored in the new databases. Older blocks would be kept in the old database.
+
+This is to ensure that we can still verify the integrity of storage after we transition to the new format. Changing the body hash of the old blocks would also change the hash of the block and hashes of all the later blocks, so it is not an option. So in order to be able to verify that the stored data still corresponds to the hashes from the blocks, we need to be able to calculate them the old way even after we start using the new way for new blocks.
+
+The code reading the blocks from storage would then have to be aware of the protocol version at which the transition happened, in order to read the block bodies from the correct databases regardless of whether it's a new or an old block.
```
goral09, line 64: Do we need specialized storage for that? If we decide that the new abstraction is only in the way we compute the hash, we could store the block as it's done right now and calculates hashes "on the fly" – pull the block by hash, calculate the Merkle proof as necessary.

text/0041-extensible-block-body.md
```diff
@@ -0,0 +1,104 @@
...
+- A `deploy_hashes` database would store the lists of deploy hashes keyed by `hash(deploy_hashes)`.
+- A `transfer_hashes` database would store the lists of transfer hashes keyed by `hash(transfer_hashes)`.
+- A `proposer` database would store the proposers keyed by `hash(proposer)`.
+
+The new way of hashing and new databases would start being used in a specific, chosen protocol version. The blocks in this protocol version and above would have their bodies hashed using the new way and be stored in the new databases. Older blocks would be kept in the old database.
+
+This is to ensure that we can still verify the integrity of storage after we transition to the new format. Changing the body hash of the old blocks would also change the hash of the block and hashes of all the later blocks, so it is not an option. So in order to be able to verify that the stored data still corresponds to the hashes from the blocks, we need to be able to calculate them the old way even after we start using the new way for new blocks.
```
goral09, line 62: Maybe the integrity verification should be version-based and we would have different pieces of logic for different versions: on node startup, during the storage integrity verification phase, we will iterate over the blocks in the storage, load each into memory, check the protocol version and verify the integrity with appropriate logic. This doesn't require changing the storage at all.

text/0041-extensible-block-body.md
fizyk20: It's also kind of about modifying how the block body is stored. But you're right, this is phrased badly.

text/0041-extensible-block-body.md
fizyk20: But I think one of the points is to be able to have incomplete blocks (@xcthulhu correct me if I'm wrong) - in such a case, the same storage format will not work.

goral09: What does "incomplete block" mean? Also, what's the use case? Can you give me an example? I work better with concrete examples.

fizyk20: By "incomplete block" I mean a block in which you downloaded and stored only some of the fields, and not all of them. So, for example, with fast sync you might want to only download the list of deploys for the purpose of checking for replays, but ignore the other fields to conserve bandwidth (the other fields are currently just the proposer, so it's not a great benefit, but we're future-proofing here :wink:).

goral09: My understanding of how fast-sync would use it is that it would download the header (it knows the correct block hash) and then request parts of that block's body (for example only transfers and hashes) – it can use the response (transfers, hashes and the merkle proof) to verify that it's indeed what it needs.

xcthulhu: > But I think one of the points is to be able to have incomplete blocks (@xcthulhu correct me if I'm wrong) - in such a case, the same storage format will not work.

To avoid replay attacks, we're going to need to download historical deploys, but that's just a fragment of a block.  We need a way to safely transmit fragments of a block.

goral09: Fragments of a block's _body_ together with the Merkle proof but my proposal also achieves that without requiring any changes from the storage layer.

goral09: > By "incomplete block" I mean a block in which you downloaded and stored only some of the fields, and not all of them. So, for example, with fast sync you might want to only download the list of deploys for the purpose of checking for replays, but ignore the other fields to conserve bandwidth (the other fields are currently just the proposer, so it's not a great benefit, but we're future-proofing here wink Or maybe I'm forgetting some other benefits).

I don't think you could consider this "incomplete block" to be a block at all. Would such node be useful for the network? Imagine a joining node trying to download the linear chain, it would query your node asking for block A only to find that you have "some" of block A and not all of it. 

Hashing the block body in the scheme you're proposing here makes sense for the extensibility purpose and for fast-sync (when it needs to download parts of the body to protect against the replay attack) but it does not respect the integrity of the data.

text/0041-extensible-block-body.md
```diff
@@ -0,0 +1,353 @@
...
+            protocol_version,
+            random_bit,
+            accumulated_seed,
+        } = self;
+
+        let hashed_era_id = hash(era_id.to_bytes());
+        let hashed_era_end = match era_end {
+            None => hash(&[]),
+            Some(era_end) => era_end.hash(),
+        };        let hashed_height = hash(height.to_bytes());
```
afck, line 246:
```suggestion
        };
        let hashed_height = hash(height.to_bytes());
```

text/0041-extensible-block-body.md
```diff
@@ -0,0 +1,353 @@
...
+        let hashed_timestamp =hash(timestamp.to_bytes());
```
afck, line 247:
```suggestion
        let hashed_timestamp = hash(timestamp.to_bytes());
```

henrytill (approved):
:+1: here

text/0041-extensible-block-body.md
```diff
@@ -0,0 +1,353 @@
...
+- A `transfer_hashes` database would store the lists of transfer hashes keyed by `hash(transfer_hashes)`.
+- A `proposer` database would store the proposers keyed by `hash(proposer)`.
+
+The new way of hashing and new databases would start being used in a specific, chosen protocol version. The blocks in this protocol version and above would have their bodies hashed using the new way and be stored in the new databases. Older blocks would be kept in the old database.
+
+This is to ensure that we can still verify the integrity of storage after we transition to the new format. Changing the body hash of the old blocks would also change the hash of the block and hashes of all the later blocks, so it is not an option. So in order to be able to verify that the stored data still corresponds to the hashes from the blocks, we need to be able to calculate them the old way even after we start using the new way for new blocks.
+
+The code reading the blocks from storage would then have to be aware of the protocol version at which the transition happened, in order to read the block bodies from the correct databases regardless of whether it's a new or an old block.
```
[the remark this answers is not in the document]
xcthulhu: It isn't clear if we will have to remember this upgrade forever.  It might be possible for non-archive nodes to move on after we have state pruning.

text/0041-extensible-block-body.md
```diff
@@ -0,0 +1,353 @@
...
+Hashing an option is similar.  Once again the sentinel hash `hash(&[])` is used to denote absence.
+
+```rust
+fn hash_option<T>(maybe_t: Option<T>) -> Result<Digest, bytesrepr::Error>
+where
+    T: ToBytes,
+{
+    match maybe_t {
+        None => Ok(hash(&[])),
+        Some(t) => Ok(hash(t.to_bytes()?)),
```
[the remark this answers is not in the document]
xcthulhu: Having discussed this, it is safer to use the zero hash (in the case of blake2bhash, `[0u8; 32]`) than it is to use `hash(&[])` throughout this.

text/0041-extensible-block-body.md
```diff
@@ -0,0 +1,357 @@
...
+```
+
+The second is a primitive for hashing slices. This follows the scheme for hashing a `BlockBody` described in the previous section. It is intended for hashing vectors where we care about order.  This is suited to hashing extendable data structures. It uses a [right fold][1].
+
+```rust
+pub fn hash_slice_rfold(slice: &[Digest]) -> Digest {
+    slice
+        .iter()
+        .rfold(SENTINEL, |prev, next| hash_pair(next, &prev))
+}
```
xcthulhu, line 103:
```suggestion
pub fn hash_slice_rfold(slice: &[Digest]) -> Digest {
    slice
        .iter()
        .rfold([2u8; HASH_SIZE], |prev, next| hash_pair(next, &prev))
}
```

As discussed, it makes it simpler to debug if there are separate sentinels for different hashing primitives.

text/0041-extensible-block-body.md
```diff
@@ -0,0 +1,357 @@
...
+g   h   i
+| /   /
+|/   /
+j   k
+| /
+|/
+l
+```
+
+Based on `hash_vec_merkle_tree`, we can give a procedure for hashing `BTreeMap`s. This procedure uses the `ToBytes` trait to abstract away they keys and values of the map.
```
Fraser999, line 160:
```suggestion
Based on `hash_vec_merkle_tree`, we can give a procedure for hashing `BTreeMap`s. This procedure uses the `ToBytes` trait to abstract away the keys and values of the map.
```

Fraser999 (commented):
I'm somewhat unconvinced that we need to go as far as what this CEP is suggesting.  Was consideration given to just handling the piece(s) of the existing `Block` in a way similar to how we already deal with this issue?  (By that I mean, in the past we split the Block body out to a separate struct and now just include a hash of this in the `BlockHeader`.)

If Fast Sync only needs the list of deploys from a block body for example, we could just split that out to a separate type, stored separately, and keep its hash in the header?

Basically, I'm not convinced the complexity and possible reduction in performance this PR introduces is worthwhile, particularly in regards to the `BlockHeader`, given that the CEP says the block header changes aren't required for Fast Sync.  I'm quite possibly missing some context though?

EdHastingsCasperLabs (approved):

EdHastingsCasperLabs:
> I'm somewhat unconvinced that we need to go as far as what this CEP is suggesting. Was consideration given to just handling the piece(s) of the existing `Block` in a way similar to how we already deal with this issue? (By that I mean, in the past we split the Block body out to a separate struct and now just include a hash of this in the `BlockHeader`.)
> 
> If Fast Sync only needs the list of deploys from a block body for example, we could just split that out to a separate type, stored separately, and keep its hash in the header?

I think this notion has merit given the current needs of the system and assuming we never actually extend the block structure further. However, I find the extensibility offered by this CEP's model compelling and IMO worth prototyping at least.

xcthulhu:
> I'm somewhat unconvinced that we need to go as far as what this CEP is suggesting. Was consideration given to just handling the piece(s) of the existing `Block` in a way similar to how we already deal with this issue? (By that I mean, in the past we split the Block body out to a separate struct and now just include a hash of this in the `BlockHeader`.)

Right now, we can't get a fragment of a `Block` and cryptographically verify it without just getting the whole `Block`. This can be problematic if we ever want to add more fields to a `Block`.  The design here is intended to allow us to flexibly add fields to `Block` while preserving our security model with clients which only need some data provided.

> If Fast Sync only needs the list of deploys from a block body for example, we could just split that out to a separate type, stored separately, and keep its hash in the header?

Currently, we have no way of doing this in a backward compatible manner.  With this CEP such things would be possible, however.  As long as a client is only looking at a fragment of a block header and a proof, then adding fields is backward compatible with that security model.

 > Basically, I'm not convinced the complexity and possible reduction in performance this PR introduces is worthwhile, particularly in regards to the `BlockHeader`, given that the CEP says the block header changes aren't required for Fast Sync. I'm quite possibly missing some context though?

It is not currently necessary for `BlockHeader`. It will allow us to potentially change the `BlockHeader` structure.  Moreover, this scheme allows for certain cryptographic proofs for clients to be more concise.  For instance, suppose there is a switch block at height `n` and I have a trusted hash at height `n+m`.  Under this scheme, to get a proof of the trusted validators from that switch block takes a little more than `m` hashes; which can more than order of magnitude less data than providing all of the headers as we must currently do.

text/0041-extensible-block-body.md
```diff
@@ -0,0 +1,378 @@
...
+        let j = k / 2 + k % 2;
+        for i in 0..j {
+            if (2 * i + 1) > k {
+                vec[i] = vec[2 * i];
+            } else {
+                vec[i] = hash_pair(&vec[2 * i], &vec[2 * i + 1]);
+            }
+        }
+        k = j;
+    }
```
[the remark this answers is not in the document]
afck: I think that's the expected behavior of a Merkle tree: It only needs a sentinel if there are 0 entries.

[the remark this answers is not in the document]
goral09: I see. Then it's worth recognizing that both folds will always use the sentinel value.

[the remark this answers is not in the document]
afck: Sure, but they're not equivalent either way: Even if the folds used no sentinel value, the Merkle tree root would still be a different hash.

PR #42, opened 2021-04-23
Network Quality-of-Service

marc-casperlabs:
[Rendered](https://github.com/marc-casperlabs/ceps/blob/network-qos/text/0042-network-qos.md)


text/0042-network-qos.md
```diff
@@ -0,0 +1,67 @@
...
+
+Every peer receives a score based on whether they are a validator (high), about to become a validator (medium) or neither (low). Peers are given guaranteed bandwidth from each according to their rank, peers scoring too low are disconnected.
+
+Low ranking peers are also subject to rotation. New connections are still accepted and connecting higher ranks may evict lower ranks.
+
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+For identification, the a peer includes its validator public key and signs its peer ID, nonced by TLS server/client random. This information is optional, but allows the networking to know that the connection belongs to a particular validator.
```
afck, line 31:
```suggestion
For identification, a peer includes its validator public key and signs its peer ID, nonced by TLS server/client random. This information is optional, but allows the networking to know that the connection belongs to a particular validator.
```

afck (approved):
Is it possible to also prioritize traffic to the same peer, based on the type of traffic? (Maybe an independent issue.)

marc-casperlabs:
> Is it possible to also prioritize traffic to the same peer, based on the type of traffic? (Maybe an independent issue.)

Yes, we added classification of networking traffic with https://github.com/casper-network/casper-node/pull/1316. This would entail inserting a priority queue or some other scheduling mechanism per node or alternative, if HTBs are used, subdividing channels even further (my preferred approach, as this would solve some starvation issues immediately).

I would tackle this as an independent issue though, as some effort needs to be put into this to avoid starvation, etc.

text/0042-network-qos.md
```diff
@@ -0,0 +1,67 @@
...
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+Without dynamic bandwidth allocation this will reduce available bandwidth at light loads (see [Future Possibilities](#future-possibilities)); strong liveness guarantees require a large number of unused channels as well.
+
+## Prior art
+
+[prior-art]: #prior-art
+
```
xcthulhu, line 46:
```suggestion
```
No need for empty sections...

text/0042-network-qos.md
```diff
@@ -0,0 +1,67 @@
...
+### Bandwidth estimates
+
+* Full blocks at 100 validators result in having to send roughly 2800 hashes at 32 bytes to each, in 33% of the round time.
+* A round is 64 seconds, leaving ~ 22 seconds for this process.
+* We include 10% margin for overhead from network encoding and TCP/IP, an almost unrealistic best case.
+* We assume a 50mbit connection (10 mbit is too small, but 100mbit across the world seems hard to guarantee).
+
+This results in a bandwidth requirement of $\frac{2800\cdot32\cdot100\cdot1.1\cdot22}{1024^2\cdot8} \approx 3.4 [\textrm{mbit}]$ for just transferring hashes and consensus, 100 KB transferred per validator. We divide by 150 slots and can transfer roughly 2.5 megabytes of additional bulk data per peer to cover gossip, address and protocol overhead, as well as bulk data sent directly. We are relying on gossiping to spread the load here, or the network will slow down.
```
xcthulhu, line 54:
```suggestion
This results in a bandwidth requirement of `2800 * 32 * 100 * 1.1 / (1024**2 * 8) ≈ 3.4` mbit for just transferring hashes and consensus, 100 KB transferred per validator. We divide by 150 slots and can transfer roughly 2.5 megabytes of additional bulk data per peer to cover gossip, address and protocol overhead, as well as bulk data sent directly. We are relying on gossiping to spread the load here, or the network will slow down.
```

xcthulhu: LaTeX isn't rendering properly on GitHub...

xcthulhu (approved):

text/0042-network-qos.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+Without dynamic bandwidth allocation this will reduce available bandwidth at light loads (see [Future Possibilities](#future-possibilities)); strong liveness guarantees require a large number of unused channels as well.
+
+## Prior art
+
+[prior-art]: #prior-art
```
goral09, line 45: You could include information about Ethereum's Go node – they also have the per-peer rating system implemented.

Fraser999 (approved):
I'm a little concerned at the `peers scoring too low are disconnected` from the perspective of the gossip protocol remaining effective, but I don't think there's much we can do about it right now, and it's outside of the scope of this CEP.

PR #43, opened 2021-04-27
Improvements to faulty validator handling

afck:
[Rendered](https://github.com/afck/ceps/blob/better_eviction/text/0043-better-eviction.md)

https://casperlabs.atlassian.net/browse/HWY-280
https://casperlabs.atlassian.net/browse/NDRS-1158

### DIFF
--- a/text/0043-better-eviction.md
+++ b/text/0043-better-eviction.md
@@ -31,7 +31,7 @@ before the new era's beginning, can choose only among eight different seeds, whi
 all likely to be fair. If they could still e.g. reduce two bids with a million tokens each, they
 could choose among a trillion different proposer sequences. Similarly, if an attacker has k
 validator nodes and could remove themselves by letting any subset of nodes exhibit a fault,
-they could choose among 2<sup>k</sup> leader sequences, even if those k nodes' stakes are small.
+they could choose among 2<sup>k</sup> proposer sequences, even if those k nodes' stakes are small.
 
 We propose a safe way to effectively ignore faulty validators anyway, even if the fault happened
 after the auction, making it easier for the network to remain live without them.
@@ -109,6 +109,23 @@ However, we believe the approach is intuitive and straightforward.
 ## Rationale and alternatives
 
 [rationale-and-alternatives]: #rationale-and-alternatives
+
+### Only modify the proposer sequence, but still count faulty validators' weights
+
+An alternative is to implement only part of this CEP: Only exclude banned validators from the
+proposer sequence, but don't set their weights to 0. This would be easier because it doesn't affect
+finality signatures and we don't have to keep track of two different sets of validators.
+
+The downside is that the faulty validators still count towards the fault tolerance threshold for the
+duration of one `auction_delay`, so the network would still cease to be live if e.g. 20% started
+being inactive in era 4, and then a further 20% started being inactive in era 5. With the full
+implementation of this CEP, such a situation would be tolerated.
+
+In either case, this can be implemented as a first step, and the full implementation can be done
+separately later.
+
+
+### Excluding faulty validators as if they hadn't won the auction
 
 It would be simpler to just remove the banned validators altogether and just apply the PRNG to the
 new weight map. However, that would reassign most slots, even those belonging to honest validators.
--- a/text/0043-better-eviction.md
+++ b/text/0043-better-eviction.md
@@ -95,7 +95,12 @@ with the banned validators removed, and use the first one only for the first ste
 assignment.
 
 Finally, the modified weights (with the banned validators set to 0) also need to be applied outside
-of Highway, wherever we count a block's finality signatures.
+of Highway, wherever we count a block's finality signatures. Any clients that validate blocks should
+use the modified weights as well.
+That suggests it would be best to put the _modified_ weights into the block header, instead of the
+original ones, or at least include all the information in every single switch block header that is
+required to compute them. The original weights would still need to be communicated to the Highway
+instance on initialization.
 
 
 ## Drawbacks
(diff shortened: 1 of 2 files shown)


text/0043-better-eviction.md
```diff
@@ -0,0 +1,126 @@
...
+everyone, so it is important that before the auction, while bids can still change, the PRNG seed is
+not yet known.
+
+After the auction, during that delay, every block adds a single bit to the random seed that
+will be used to initialize the PRNG. That way, an attacker, even if they had the last three blocks
+before the new era's beginning, can choose only among eight different seeds, which are
+all likely to be fair. If they could still e.g. reduce two bids with a million tokens each, they
+could choose among a trillion different proposer sequences. Similarly, if an attacker has k
+validator nodes and could remove themselves by letting any subset of nodes exhibit a fault,
+they could choose among 2<sup>k</sup> leader sequences, even if those k nodes' stakes are small.
```
xcthulhu, line 34:
```suggestion
they could choose among 2ᵏ leader sequences, even if those k nodes' stakes are small.
```

afck: It does render correctly on GitHub, though.

xcthulhu (approved):

text/0043-better-eviction.md
```diff
@@ -0,0 +1,126 @@
...
+They only lose some of their own slots.
+
+
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+The first change to the protocol is that in addition to the equivocators, anyone who was reported
+as inactive in a switch block since the auction is marked as banned as well, when a new era is
+initialized. We will apply this only the the past `auction_delay` eras: If someone exhibited a fault
```
goral09, line 68:
```suggestion
initialized. We will apply this only to the past `auction_delay` eras: If someone exhibited a fault
```

goral09 (approved):

text/0043-better-eviction.md
```diff
@@ -0,0 +1,126 @@
...
+earlier than that, the auction contract will already have removed them.
+
+The second change is a modification of the proposer slot assignment:
+* We currently add (wrapping, as addition of two unsigned 64-bit integers) the era's seed to the
+  slot's timestamp and seed a
+  [ChaCha PRNG](https://docs.rs/rand_chacha/0.3.0/rand_chacha/struct.ChaCha8Rng.html)
+  with the result, then apply it to select a validator, weighted by their stakes.
+* Now, if the result is a banned validator, we use the same seed, but select a different validator,
```
AlexanderLimonov, line 76: It's a bit unclear what this means. Do you suggest advancing the sequence until you hit a valid entry, or something else?

text/0043-better-eviction.md
```diff
@@ -0,0 +1,126 @@
...
+  choosing only among the non-banned ones, weighted by only _their_ stakes.
+
+Thirdly, we remove the banned validators' weights from the total weight for the purpose of fork
```
AlexanderLimonov, line 79: It seems that the proposal here is to, _de facto_, just treat inactive validators as equivocators (without slashing, if it's activated). Is that the case? If so, maybe it would be useful to add an exhaustive description of that case and make it clear that we're just merging the cases.

text/0043-better-eviction.md
```diff
@@ -0,0 +1,126 @@
...
+pseudorandomly. It is important that an attacker cannot manipulate the system in a way that allows
+them to choose among a large number of proposer sequences and thus get more slots. That is why
+the validator weights are determined in advance, with an `auction_delay` between the auction and
+the beginning of the era: A small change to any weight would change the proposer sequence for
+everyone, so it is important that before the auction, while bids can still change, the PRNG seed is
+not yet known.
+
+After the auction, during that delay, every block adds a single bit to the random seed that
+will be used to initialize the PRNG. That way, an attacker, even if they had the last three blocks
+before the new era's beginning, can choose only among eight different seeds, which are
```
AlexanderLimonov, line 30: This isn't a comment on the CEP as such, but this business with strategically selecting seeds seems like something that could use a simulation.

text/0043-better-eviction.md
```diff
@@ -0,0 +1,126 @@
...
+
+In Highway, validators take turns at proposing blocks, and the sequence of proposers is determined
+pseudorandomly. It is important that an attacker cannot manipulate the system in a way that allows
+them to choose among a large number of proposer sequences and thus get more slots. That is why
+the validator weights are determined in advance, with an `auction_delay` between the auction and
+the beginning of the era: A small change to any weight would change the proposer sequence for
+everyone, so it is important that before the auction, while bids can still change, the PRNG seed is
+not yet known.
+
+After the auction, during that delay, every block adds a single bit to the random seed that
```
AlexanderLimonov, line 28: Maybe this could use an extended example. When does the malicious validator go inactive, and which blocks is he trying to manipulate, and in what era?

text/0043-better-eviction.md
afck: This is only on the consensus side, for the (in mainnet) single era in between, where the inactivity has already been detected but not affected the official validator set as determined by the auction contract yet. This CEP isn't meant to make any recommendation about what to do with equivocators vs. inactive validators later; it just allows us to ignore them before until the auction contract officially removes them.

Because if a validator was completely silent in era 5, we shouldn't expect it to contribute again in era 6; it's safer to remove it from the validator set for fork choice and finality purposes, so the network does not depend on it for liveness or safety at all anymore.

text/0043-better-eviction.md
afck: Done in https://github.com/casper-network/ceps/pull/43/commits/b31202ed858d95fd7338741c07cf45582dd4dbaa.

text/0043-better-eviction.md
afck: I added an example in https://github.com/casper-network/ceps/pull/43/commits/b31202ed858d95fd7338741c07cf45582dd4dbaa.

text/0043-better-eviction.md
afck: I added an example in https://github.com/casper-network/ceps/pull/43/commits/b31202ed858d95fd7338741c07cf45582dd4dbaa, in the "Rationale and alternatives" section.

goral09 (approved):

PR #47, opened 2021-05-13
CEP47 - Casper NFT Protocol

zie1ony:
[rendered](https://github.com/casper-network/ceps/blob/cep46_casper_nft_protocol/text/0047-casper-nft-protocol.md)


goral09:
Can you change the file to a Markdown format?

EdHastingsCasperLabs (approved):
Implementations of this have been in use for months now; thus it is de facto accepted. 

Related, we are drafting a new CEP for NFT's, a reference implementation, client support, etc. which incorporates newer features that have been added to the protocol; which work should be completed by the end of Q1 2022.

piotr-dziubecki (approved):

PR #48, opened 2021-05-14
CEP 48: New Synchronizer

afck:
[Rendered](https://github.com/afck/ceps/blob/sync/text/0048-synchronizer-redesign.md)

https://github.com/casper-network/casper-node/issues/1426

### DIFF
--- a/text/0048-synchronizer-redesign.md
+++ b/text/0048-synchronizer-redesign.md
@@ -27,10 +27,6 @@ The synchronizer's job is to decide when to request dependencies, and from which
 
 The current implementation started off as a naive way of doing this, and has received some isolated improvements, but still makes redundant requests sometimes and is not as efficient as it could be.
 
-Also, the protocol overhead and even its asymptotic best-case complexity could be reduced considerably by replacing the panoramas — the lists of validators' latest unit hashes in each unit — with compressed lists of unit sequence numbers and a single checksum. Using sequence numbers instead of hashes changes the synchronizer logic considerably. This idea is based on section A.3 of [_Aleph: Efficient Atomic Broadcast in Asynchronous Networks with Byzantine Nodes_][Aleph].
-
-[Aleph]: https://arxiv.org/pdf/1908.05156.pdf
-
 
 ## Guide-level explanation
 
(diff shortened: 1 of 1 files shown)


text/0048-synchronizer-redesign.md
```diff
@@ -0,0 +1,170 @@
...
+
+### New dependency type: "Unit with sequence number"
+
+In addition to specifying units by hash, we add a message type
+
+```rust
+UnitBySeqNumber(u64, Hash)
+```
+
+which means we ask for a unit with the specified sequence number by the same sender as the unit with the given hash.
```
goral09, line 66: What if the sender gives us a unit with the correct sequence number but a different parent? i.e. a unit from a different fork. This will make the node start downloading units from a different chain, back to the beginning of an era.

text/0048-synchronizer-redesign.md
```diff
@@ -0,0 +1,170 @@
...
+Finally, process the queue.
+
+
+### Processing the queue
+
+Iterate over all pending units, sorted by their timestamp, up to the current timestamp, twice:
+
+In the first pass:
+
+* Check whether the unit depends on evidence we don't have, and request it if possible.
```
goral09, line 99: _evidence_ specifically or an type of dependency?

goral09: Unless you meant to download unit-type dependencies in previous steps and evidence at the last stage since satisfying an evidence request is different from a unit dependency.

text/0048-synchronizer-redesign.md
afck: In the first step of the second queue processing pass it says we use this only if "the unit's creator is not known to be faulty". So what will happen if Alice created `a0, a1, a2, a3, a4`, but also a fork `a0', a1', a2', a3'`, and we have `a0, a1`, and she sends us `a4'`? We'll request "unit number 2", since our latest one is `a1`. If she sends us `a2'`, that will be earlier in the queue, since it has an earlier timestamp, so we'll _first_ try to request `a2'`'s dependencies. That includes `a1'`, which we'd request _by hash_. But as soon as we have `a1'` and `a1`, we know that Alice is faulty, so we will stop requesting units by sequence number from her.

goral09 (approved):

text/0048-synchronizer-redesign.md
```diff
@@ -0,0 +1,170 @@
...
+
+To keep the synchronizer's memory usage small and its computations fast, we want to keep its state as small as possible, so we want to prioritize vertices in a way that minimizes the time we expect them to remain pending in the synchronizer: We always want to request those vertices first which are most likely to be added to the protocol state soon. That roughly means: first all evidence against faulty validators, since evidence never has a dependency and can always be added to the protocol state right away; then all endorsements and units, starting with those with an early timestamp, since they tend to have fewer dependencies.
+
+Based on that prioritization we want to send requests so that:
+
+* We request each vertex from only one peer at a time, but with an aggressive timeout.
+* We don't request too many vertices from the same peer at once. The number will be configurable. A low default — even 1 or 2 — should be fine for a running node, but we should determine experimentally whether a higher value speeds up syncing for a joining node.
```
marc-casperlabs, line 48: I'm not sure about the 1-2 node limit, if it prevents us from queuing all the necessary fetches, we will introduce a round-trip worth of latency to the process each time.

As a first step, we could adjust the weight to get an even distribution, and when a request fails, we redistribute all outstanding ones to other peers?

Another idea is to support batching, requesting more than one vertex at the same time from a single peer and supporting multiple in the response. The peer would reply with both vertices and acknowledges that some are missing, and is allowed to do multiple replies. We would need to keep the message size reasonable though.

Finally is it possible to describe a whole subgraph through one missing vertex maybe? Some like a "get vertex V and all of its dependencies (up to a depth/deliminated by some known hash I am providing as well)"?

All in all, can we lower the number of roundtrips on this by either requesting more in parallel, batching and/or letting the sender figure out what we need?

afck: As discussed, sending multiple requests while others are still in flight has roughly the same effect as batching, on the network level. We'll make the limit for parallel requests to the same peer configurable, so we can test with a high value.

> describe a whole subgraph through one missing vertex maybe

I'd like to leave that for later, as a possible improvement: Getting this right and considering all edge cases (equivocations, endorsements, …) will be pretty complicated. Also, if we request too much at once, the probability increases that another peer will send us some of the same information in the meantime.

PR #55, opened 2021-06-08
CEP: Signature Based Finality

### LINKED ISSUE #54: Create a CEP for signature based finality
Signature based finality is a prerequisite for consensus memory footprint optimizations. Due to the extent of changes required, a CEP is needed for them.

fizyk20:
This elaborates on details of a part of #36.

[Rendered](https://github.com/fizyk20/ceps/blob/signature-based-finality/text/0055-signature-based-finality.md)

Closes #54.

### DIFF
--- a/text/0055-signature-based-finality.md
+++ b/text/0055-signature-based-finality.md
@@ -39,22 +39,26 @@ In order to only consider fully signed blocks as finalized, we'll have to change
 3. Contract runtime executes the block and announces a `Block`.
 4. In response to the announcement, linear chain component caches the block, and consensus creates a finality signature.
 5. Linear chain stores the signature and gossips it. It also receives signatures sent over the network.
-6. Once the linear chain gathers enough signatures for the block, it considers it finalized.
+6. Once the linear chain collects enough signatures for the block, it considers it finalized.
     - If the block is unknown yet, it requests the block from the sender of the last signature.
-    - If the block is known, but it isn't the immediate descendant of the highest block so far, it requests the range of blocks from the immediate descendant of the highest block to the just finalized block.
+    - If the block is known, but it isn't the immediate descendant of the highest block stored so far, it requests the range of blocks from the immediate descendant of the highest block to the just finalized block.
     - If the block is known and it is the immediate descendant of the highest block so far, it adds the block to the chain as the new highest block.
 
+Note: with this process, if a block is stored in the linear chain, it also implies that it is known (that is, we know not just its hash, but all its contents), finalized by consensus and fully signed, because we will only store blocks after they meet all these criteria.
+
 A node receiving the request sent in point 6 would respond with the following:
+
 - To the request for a single fully signed block, send the block along with signatures proving it is fully signed.
 - To the request for a range of fully signed blocks, send the fully signed blocks one by one, starting from the oldest one.
 
 Upon receiving a block with signatures as a response:
-    - If it's not fully signed, discard and re-request it from someone else.
-    - If it's the immediate descendant of the highest known block:
-        - Request execution by the contract runtime.
-        - Compare the execution results to the block - if they don't match, discard the block and re-request it from someone else.
-        - If everything fits, store the block as the new highest block.
-    - If it's not the immediate descendant of the highest known block, request the range of blocks from the immediate descendant of the highest block to this one.
+
+- If it's not fully signed, cache the signatures received and re-request it from someone else.
+- If it's the immediate descendant of the highest known block:
+    - Request execution by the contract runtime.
+    - Compare the execution results to the block - if they don't match, print an error and shut down the node. This would mean that either the node software has a major bug, or the network was taken over by malicious actors.
+    - If everything fits, store the block as the new highest block.
+- If it's not the immediate descendant of the highest known block, request the range of blocks from the immediate descendant of the highest block to this one.
 
 ## Reference-level explanation
 
--- a/text/0055-signature-based-finality.md
+++ b/text/0055-signature-based-finality.md
@@ -32,43 +32,42 @@ The current process of block finalization is this:
 5. Consensus creates a finality signature for the block.
 6. Linear chain stores the signature and gossips it.
 
-In order to only consider fully signed blocks as finalized, we'll have to change this process to the following:
+We wouldn't change the meaning of "finalized" in order not to introduce confusion - a block being finalized would still mean finalized by consensus. However, we would add a new, "signed" status of the block, which would be used instead of "finalized" for some purposes (for example, advancing to the next era will require the switch block to be signed, instead of only finalized). The process would then look the following way:
 
 1. A `BlockPayload` gets finalized by a consensus instance.
 2. The consensus component announces a `FinalizedBlock`.
 3. Contract runtime executes the block and announces a `Block`.
-4. In response to the announcement, linear chain component caches the block, and consensus creates a finality signature.
-5. Linear chain stores the signature and gossips it. It also receives signatures sent over the network.
-6. Once the linear chain collects enough signatures for the block, it considers it finalized.
+4. In response to the announcement, linear chain component caches the block and announces that it was added.
+5. Consensus creates a finality signature for the block.
+6. Linear chain stores the signature and gossips it. It also receives signatures sent over the network.
+7. Once the linear chain collects enough signatures for the block, the block is considered signed.
     - If the block is unknown yet, it requests the block from the sender of the last signature.
-    - If the block is known, but it isn't the immediate descendant of the highest block stored so far, it requests the range of blocks from the immediate descendant of the highest block to the just finalized block.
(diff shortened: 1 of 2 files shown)


afck (approved):

text/0055-signature-based-finality.md
```diff
@@ -0,0 +1,115 @@
...
+6. Linear chain stores the signature and gossips it.
+
+In order to only consider fully signed blocks as finalized, we'll have to change this process to the following:
+
+1. A `BlockPayload` gets finalized by a consensus instance.
+2. The consensus component announces a `FinalizedBlock`.
+3. Contract runtime executes the block and announces a `Block`.
+4. In response to the announcement, linear chain component caches the block, and consensus creates a finality signature.
+5. Linear chain stores the signature and gossips it. It also receives signatures sent over the network.
+6. Once the linear chain gathers enough signatures for the block, it considers it finalized.
```
goral09, line 42:
```suggestion
6. Once the linear chain collects enough signatures for the block, it considers it finalized.
```

text/0055-signature-based-finality.md
```diff
@@ -0,0 +1,115 @@
...
+    - If the block is unknown yet, it requests the block from the sender of the last signature.
+    - If the block is known, but it isn't the immediate descendant of the highest block so far, it requests the range of blocks from the immediate descendant of the highest block to the just finalized block.
```
goral09, line 44: I'm guessing that the _the highest block so far_ means highest _known_ and _stored_ and _finalized_?

goral09: Who does it request this range of blocks from? The sender of the last signature as well?

text/0055-signature-based-finality.md
```diff
@@ -0,0 +1,115 @@
...
+- To the request for a single fully signed block, send the block along with signatures proving it is fully signed.
+- To the request for a range of fully signed blocks, send the fully signed blocks one by one, starting from the oldest one.
+
+Upon receiving a block with signatures as a response:
+    - If it's not fully signed, discard and re-request it from someone else.
+    - If it's the immediate descendant of the highest known block:
+        - Request execution by the contract runtime.
+        - Compare the execution results to the block - if they don't match, discard the block and re-request it from someone else.
+        - If everything fits, store the block as the new highest block.
+    - If it's not the immediate descendant of the highest known block, request the range of blocks from the immediate descendant of the highest block to this one.
```
goral09, line 57: This part doesn't render correctly as a bullet list for me.

text/0055-signature-based-finality.md
```diff
@@ -0,0 +1,115 @@
...
+    - If the block is unknown yet, it requests the block from the sender of the last signature.
+    - If the block is known, but it isn't the immediate descendant of the highest block so far, it requests the range of blocks from the immediate descendant of the highest block to the just finalized block.
+    - If the block is known and it is the immediate descendant of the highest block so far, it adds the block to the chain as the new highest block.
+
+A node receiving the request sent in point 6 would respond with the following:
+- To the request for a single fully signed block, send the block along with signatures proving it is fully signed.
+- To the request for a range of fully signed blocks, send the fully signed blocks one by one, starting from the oldest one.
+
+Upon receiving a block with signatures as a response:
+    - If it's not fully signed, discard and re-request it from someone else.
```
goral09, line 52: Can we include here, as future steps at least, that we should be collecting "incomplete" signatures? It would be better for performance if we didn't expect a single source to have all the signatures immediately.

text/0055-signature-based-finality.md
```diff
@@ -0,0 +1,115 @@
...
+    - If it's the immediate descendant of the highest known block:
+        - Request execution by the contract runtime.
+        - Compare the execution results to the block - if they don't match, discard the block and re-request it from someone else.
```
goral09, line 55: If we're here, it means we received a block that is an immediate descendant of our last known/finalized/stored block, and the block we have just received has >2/3 of finality signatures. Should we really expect that other nodes will give us a block with the different post-execution results? Even if they did, this can't possibly sum up to the required 2/3 since we had just rejected a block signed by the quorum this big. If it does though, then it means there was an equivocation.

I think we should exit the node if the execution results of a block signed by more than 2/3 validators don't match.

cc @afck @xcthulhu

text/0055-signature-based-finality.md
goral09, line 57: Are we going to use the (already-existing) `Fetcher` implementation for this?

text/0055-signature-based-finality.md
afck: That's a good point, I agree.

text/0055-signature-based-finality.md
```diff
@@ -0,0 +1,115 @@
...
+
+`RequestBlockRange(a, b)` handles the use case when a new fully signed block is not the immediate descendant of the highest known fully signed block. `a` and `b` denote the minimum and maximum requested block heights.
+
+`FullySignedBlock(block, signatures)` would be used as a response to a `RequestBlockRange`. Sometimes, multiple such responses would be required.
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+This complicates the process of finalization slightly, and will probably increase time to finality (the time elapsed between when the block was proposed and when it was fully finalized, ie. fully signed). However, the significant memory savings enabled by this are worth the tradeoff.
```
goral09, line 83: I am not sure it would increase the _time-to-finality_ since, from the perspective of the single node the block is finalized as soon as that node detects the finality. When will we produce an event to the event stream about block's finalization now?

goral09: I see now, in the _Reference-level implementation_:

> Only after it is fully signed and the immediate ancestor of the highest finalized block, will it be appended to the linear chain and announced in a `BlockAdded` announcement.

text/0055-signature-based-finality.md
fizyk20: Highest stored in the linear chain. After this proposal, it will only get stored once it is known, finalized by consensus and fully signed.

I'll clarify in the text.

goral09 (approved):
Looks good in general, except for the small issue mentioned in [this comment](https://github.com/casper-network/ceps/pull/55/files#r648349348).

text/0055-signature-based-finality.md
fizyk20: I was going to say that the source should have them, because we'll only request fully signed blocks _after_ a descendant of it is fully signed, but of course we might see a block as fully signed while some of the validators who signed it still don't. Good point :+1:

text/0055-signature-based-finality.md
fizyk20: I was thinking of a separate implementation, but maybe the `Fetcher` is a good fit here. I'll have to take a closer look.

goral09: That's fine but I just want to note that this isn't as straightforward to get right as it may seem. You have to remember about:
* de-duplicating requests for the same block / to the same node
* about request timeouts
* about throttling (to not spam a single node)
to name a few.

text/0055-signature-based-finality.md
```diff
@@ -32,43 +32,42 @@ The current process of block finalization is this:
 5. Consensus creates a finality signature for the block.
 6. Linear chain stores the signature and gossips it.
 
-In order to only consider fully signed blocks as finalized, we'll have to change this process to the following:
+We wouldn't change the meaning of "finalized" in order not to introduce confusion - a block being finalized would still mean finalized by consensus. However, we would add a new, "signed" status of the block, which would be used instead of "finalized" for some purposes (for example, advancing to the next era will require the switch block to be signed, instead of only finalized). The process would then look the following way:
```
afck, line 35: Maybe add that the meaning of the "signed" status is that we have a _concise_ proof of finality. (As opposed to the complicated, large one, that would be a Highway summit.)

text/0055-signature-based-finality.md
```diff
@@ -32,43 +32,42 @@ The current process of block finalization is this:
...
 2. The consensus component announces a `FinalizedBlock`.
 3. Contract runtime executes the block and announces a `Block`.
-4. In response to the announcement, linear chain component caches the block, and consensus creates a finality signature.
-5. Linear chain stores the signature and gossips it. It also receives signatures sent over the network.
-6. Once the linear chain collects enough signatures for the block, it considers it finalized.
+4. In response to the announcement, linear chain component caches the block and announces that it was added.
+5. Consensus creates a finality signature for the block.
+6. Linear chain stores the signature and gossips it. It also receives signatures sent over the network.
+7. Once the linear chain collects enough signatures for the block, the block is considered signed.
     - If the block is unknown yet, it requests the block from the sender of the last signature.
```
afck, line 44: And after a timeout, I guess we'd request it from the sender of the next signature and so on? It should probably work in a similar way to how the block validator downloads deploys.

afck (approved):

PR #56, opened 2021-06-08
CEP 56: Sequence number panoramas

afck:
[Rendered](https://github.com/afck/ceps/blob/seq_num_panorama/text/0056-sequence-number-panorama.md)

Closes https://github.com/casper-network/ceps/issues/53.

### DIFF
--- a/text/0056-sequence-number-panorama.md
+++ b/text/0056-sequence-number-panorama.md
@@ -1,0 +1,153 @@
+# Sequence number panoramas
+
+## Summary
+
+[summary]: #summary
+
+CEP PR: [casperlabs/ceps#0056](https://github.com/casperlabs/ceps/pull/0056)
+
+We propose citing units by sequence number instead of by hash.
+This will reduce the protocol overhead and bandwidth requirement.
+
+
+## Motivation
+
+[motivation]: #motivation
+
+32-byte hashes of earlier units make up the largest part of Highway messages.
+Replacing them by 8-byte sequence numbers will therefore considerably reduce message sizes,
+and improve performance.
+
+
+## Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+Each unit — the main type of message in the Highway protocol — contains a list of hashes of earlier units.
+That way the sender attests to which units they had received before they created the new one.
+This is what defines the directed acyclic graph (DAG) structure of Highway's protocol state:
+The graph has an arrow `u → v` if `u` _directly cites_ `v`, i.e. the message `u` mentions `v` by hash.
+
+What matters to the protocol's logic is not the graph itself but the partial order it defines:
+If `u → v → w` that means the creator of `u` must also have known about `w`, even if `u` doesn't directly cite `w`.
+We write `u > v` if `v` is reachable from `u` by following any number of arrows.
+
+In particular, the actual message `u` that gets sent over the wire does not need to _directly_ cite every `v < u`.
+In the current implementation, `u` contains a _panorama_, which is a list containing for each (honest) validator the hash of the latest `v < u` created by that validator.
+
+We propose replacing the hash with the _sequence number_ instead, i.e. the number of earlier units by the same validator.
+If Alice has created units `a0`, `a1` and `a2` so far, and Bob wants to cite `a2`, instead of its hash, he will now write `2` in the panorama of his unit.
+
+[Panorama]: https://github.com/casper-network/casper-node/blob/b56c29cc97e86ead8d36dfdb2d41a8ecb1dffce9/node/src/components/consensus/highway_core/state/panorama.rs#L25
+
+
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+Units in memory still contain the full hash-based panorama to allow fast lookups.
+
+But before sending a unit `u` over the wire we replace each hash referring to a `v < u` with `v`'s sequence number.
+In addition, we add the hash of `u`'s full hash-based panorama so the recipient can verify after recomputing it.
+And we add `u`'s own predecessor by hash, i.e. the previous unit by the same creator.
+
+The unit wire format now looks like this (the first three fields are new, and the `panorama` field is removed):
+
+```rust
+struct WireUnit {
+    seq_num_panorama: Vec<ObservationSeqNum>,
+    panorama_hash: Hash,
+    previous: Option<Hash>,
+    creator: ValidatorIndex,
+    instance_id: InstanceId,
+    value: Option<BlockPayload>,
+    seq_number: u64,
+    timestamp: Timestamp,
+    round_exp: u8,
+    endorsed: BTreeSet<Hash>,
+}
+
+enum ObservationSeqNum {
+    Correct(u64),
+    Faulty,
+    None,
+}
+```
+
+`None` means that we haven't seen any unit from that validator yet, `Faulty` means they have equivocated.
+`Correct(n)` is the direct citation of the validator's `1 + n`-th unit.
+
+Upon receiving a unit, we request all missing dependencies from the sender.
+This is now a bit more complicated than before, since we cannot request all units by hash anymore:
+
+* As long as we don't have units with the required sequence numbers we request the missing ones, by sequence number.
+* For validators with the entry `Faulty` we request evidence unless we already have it.
+* If/when we have all units with the required sequence numbers, we use their hashes to reconstruct the hash-based panorama.
+* If the reconstructed panorama's hash does not match the unit's `panorama_hash`, or if any validator is cited as `Correct` that we know to be faulty, we request the full hash-based panorama from the sender.
+* Once we have the hash-based panorama, we request any missing units by hash, as well as endorsements (see section 3.6.1 of the [_Highway_ paper][Highway]) mentioned in the unit.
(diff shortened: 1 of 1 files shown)


text/0056-sequence-number-panorama.md
```diff
@@ -0,0 +1,153 @@
...
+
+enum ObservationSeqNum {
+    Correct(u64),
+    Faulty,
+    None,
+}
+```
+
+`None` means that we haven't seen any unit from that validator yet, `Faulty` means they have equivocated.
+`Correct(n)` is the direct citation of the validator's `1 + n`-th unit.
```
fizyk20, line 78: Why `1+n`-th, and not `n`-th here?

afck: Because we start counting at `0`, so the unit with sequence number `n` is the `1 + n`-th unit.

fizyk20: Ah, I see! It is slightly confusing, though - maybe something like "the validator's unit with sequence number `n`" would be better, instead?

text/0056-sequence-number-panorama.md
```diff
@@ -0,0 +1,153 @@
...
+What matters to the protocol's logic is not the graph itself but the partial order it defines:
+If `u → v → w` that means the creator of `u` must also have known about `w`, even if `u` doesn't directly cite `w`.
+We write `u > v` if `v` is reachable from `u` by following any number of arrows.
+
+In particular, the actual message `u` that gets sent over the wire does not need to _directly_ cite every `v < u`.
+In the current implementation, `u` contains a _panorama_, which is a list containing for each (honest) validator the hash of the latest `v < u` created by that validator.
+
+We propose replacing the hash with the _sequence number_ instead, i.e. the number of earlier units by the same validator.
+If Alice has created units `a0`, `a1` and `a2` so far, and Bob wants to cite `a2`, instead of its hash, he will now write `2` in the panorama of his unit.
+
```
xcthulhu, line 40: We should add here that we fall back to the legacy system in the event of an equivocation.

text/0056-sequence-number-panorama.md
```diff
@@ -0,0 +1,153 @@
...
+That way the sender attests to which units they had received before they created the new one.
+This is what defines the directed acyclic graph (DAG) structure of Highway's protocol state:
+The graph has an arrow `u → v` if `u` _directly cites_ `v`, i.e. the message `u` mentions `v` by hash.
+
+What matters to the protocol's logic is not the graph itself but the partial order it defines:
+If `u → v → w` that means the creator of `u` must also have known about `w`, even if `u` doesn't directly cite `w`.
+We write `u > v` if `v` is reachable from `u` by following any number of arrows.
+
+In particular, the actual message `u` that gets sent over the wire does not need to _directly_ cite every `v < u`.
+In the current implementation, `u` contains a _panorama_, which is a list containing for each (honest) validator the hash of the latest `v < u` created by that validator.
```
xcthulhu, line 36:
```suggestion
In the current implementation, `u` contains a _panorama_, which is a list containing for each (non-faulty) validator the hash of the latest `v < u` created by that validator.
```

text/0056-sequence-number-panorama.md
```diff
@@ -0,0 +1,153 @@
...
+
```
xcthulhu, line 37: Perhaps mention that for non-faulty validators, for units `u` and `v` that they created we have either `u > v` or `u < v` or `u = v`.  Hence, units form a *sequence* and can be assigned sequence numbers.  Then go on to the next paragraph below to talk about how we can use this domain knowledge for the optimization proposed.

text/0056-sequence-number-panorama.md
```diff
@@ -0,0 +1,153 @@
...
+32-byte hashes of earlier units make up the largest part of Highway messages.
+Replacing them by 8-byte sequence numbers will therefore considerably reduce message sizes,
+and improve performance.
+
+
+## Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+Each unit — the main type of message in the Highway protocol — contains a list of hashes of earlier units.
```
xcthulhu, line 26:
```suggestion
Each unit — the main type of message in the Highway protocol — is a hash of hashes of earlier units.
```

text/0056-sequence-number-panorama.md
```diff
@@ -0,0 +1,153 @@
...
+That way the sender attests to which units they had received before they created the new one.
+This is what defines the directed acyclic graph (DAG) structure of Highway's protocol state:
+The graph has an arrow `u → v` if `u` _directly cites_ `v`, i.e. the message `u` mentions `v` by hash.
```
xcthulhu, line 29:
```suggestion
The graph has an arrow `u → v` if `u` _directly cites_ `v`, i.e. the hash of `v` was used to generate `u`'s hash.
```

text/0056-sequence-number-panorama.md
afck: But a unit is not a hash. It really is a structure _containing_ a panorama, i.e. a list of hashes.

text/0056-sequence-number-panorama.md
afck: But it's not really about `u`'s hash, is it? It's about whether `u` attests to `v`, by mentioning `v` by hash. (Currently — or by sequence number and control hash after this CEP.)

fizyk20: Hmm... But it might be a good point - the hash of a unit should still be calculated using the hash-based panorama, right? Otherwise we lose the guarantee that the hash of a unit uniquely identifies both a unit _and_ its ancestors. Is it still calculated this way in your PR? I remember you changed some fields in the unit struct there.

Maybe this is actually worth stressing somewhere in the CEP? Not necessarily here, though, because I agree that the point raised here isn't about hashes, it's more of an expression of an abstract property.

fizyk20: Ah, the unit still contains the hash of the (hash-based) panorama, so it should be fine :+1:

afck: Ah, I see, yes, maybe that's what @xcthulhu means?

text/0056-sequence-number-panorama.md
```diff
@@ -0,0 +1,156 @@
...
+
+What matters to the protocol's logic is not the graph itself but the partial order it defines:
+If `u → v → w` that means the creator of `u` must also have known about `w`, even if `u` doesn't directly cite `w`.
+We write `u > v` if `v` is reachable from `u` by following any number of arrows.
+
+In particular, the actual message `u` that gets sent over the wire does not need to _directly_ cite every `v < u`.
```
goral09, line 35: Can we be consistent with the direction of the arrows and signs? Previously, it was `u > v`, now it's `v < u`, I know it's the same but it would be simpler, in my opinion, if we stick to one.

afck: Done in https://github.com/casper-network/ceps/pull/56/commits/8afa9eba4a570747f75f7215ae29949f8356dd96.

text/0056-sequence-number-panorama.md
afck: Done in https://github.com/casper-network/ceps/pull/56/commits/8afa9eba4a570747f75f7215ae29949f8356dd96.

text/0056-sequence-number-panorama.md
afck: Done in https://github.com/casper-network/ceps/pull/56/commits/365531c9b716b67564fa5c7a74ff6698d4833178.

text/0056-sequence-number-panorama.md
afck: Done in https://github.com/casper-network/ceps/pull/56/commits/d8954d134ac89000dc2a12398e61fdff68ea0663.

text/0056-sequence-number-panorama.md
afck: Done in https://github.com/casper-network/ceps/pull/56/commits/d8954d134ac89000dc2a12398e61fdff68ea0663.

text/0056-sequence-number-panorama.md
xcthulhu: My mistake!

text/0056-sequence-number-panorama.md
xcthulhu: > the hash of a unit should still be calculated using the hash-based panorama, right?

@fizyk20 This is what I am trying to express, correct.

It's crucial to emphasize that cryptographic integrity does not rely on sequence numbers alone, but hashes of directly cited units.  A directly cited hash uses its sequence number as part of how the hash is generated.

Switching to sequence numbers is basically switching to an index.  Most of the time nodes will not be faulty.  While none of the nodes are faulty we can optimistically use indices.  When the system is faulty, we need to fall back to the slow but bulletproof hash-based approach of getting units from peers.

afck: OK, I'll add that to the text. I'd like to do it in a different line, though, since this one is about defining the order itself, and I'd like to separate that from the cryptographic integrity question.

xcthulhu (approved):

goral09 (approved):

afck:
bors merge

PR #61, opened 2021-07-09
Decouple delegation rate changes from auction delay

Staking7pc:


Staking7pc:
I think I missed out on the main bit on how we can restrict the add-bid. wasm if the bid was already in place and allow a new wasm which will ensure the changes described above is reflected.

sacherjj:
This is a matter of trust. While a single Era delay on delegation rate is a method that a validator could operate unexpectantly, I'm not sure that restriction of the protocol is the best fix for this. Your drawbacks is quite light and does not include the massive change of auction functionality in a moving network.  So this is something not done lightly and all solutions to mitigate this issue should be considered.

In the simplest form, delegation is from trust.

- Trust that validator will run their node in a stable form and not be ejected (dropping delegator rewards to zero).
- Trust that their delegation rate will change slowly or change is communicated.
- Trust that their validator will take upgrades in time, which is a special case of number 1.

A better location for this is tracking uptime rate and large delegation rate changes as a validator measure. This would work well with cspr.live validator registration. A validator that is a bad actor loses faith. If this is recorded and tracked, then it can be remembered.  Lost trust takes a long time to rebuild.  So a short term gain would provide a long term loss for the validator and it does not make sense to behave in that manner.

MParlikar:
Can you please provided a rendered view?   See closed, merged CEP's to see the patterns.

piotr-dziubecki:
We are taking it under consideration for 2.0

RitaMAllenCA:
@deuszex Hi - this looks related to the work you are doing regarding the sandwich attacks - perhaps a standard or a PR that needs to be closed?  Please take a look and follow up in a standup this week - or via slack - thank you

devendran-m:
Won't do.

PR #66, opened 2021-09-07
Archival nodes

### LINKED ISSUE #62: Gap filling Design
We need a design for an archival node.  This way historical data from the blockchain can be preserved.

fizyk20:
[Rendered.](https://github.com/fizyk20/ceps/blob/archival-nodes/text/0000-archival-nodes.md)

Closes #62

### DIFF
--- a/text/0000-archival-nodes.md
+++ b/text/0000-archival-nodes.md
@@ -1,0 +1,86 @@
+# Archival Nodes
+
+## Summary
+
+[summary]: #summary
+
+CEP PR: [casperlabs/ceps#0000](https://github.com/casperlabs/ceps/pull/0000)
+
+This CEP outlines the design of archival nodes, that is, nodes keeping the complete history of the Casper blockchain.
+
+## Motivation
+
+[motivation]: #motivation
+
+The Casper blockchain, like all blockchains, is rapidly growing, and requires more and more storage space to be kept in its entirety. This is very wasteful, as only a small number of most recent blocks are strictly required for regular operations. It is thus tempting to simply discard the blocks that are no longer relevant, but this would mean losing almost all of the network's history, which could turn out to be catastrophic if something goes wrong, and would go against the idea of a blockchain being a public ledger of all operations ever executed on it.
+
+There is a middle ground here, though - instead of having all nodes keep the whole history, or having all nodes discard older blocks, we could have most of the nodes only keeping what is relevant for day-to-day operations, and only a small number of nodes keeping the complete history of the network. These history keepers would be called the "archival nodes".
+
+## Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+Until Fast Sync is introduced, all nodes are effectively archival nodes, as every time a node joins the network, it has to sync the whole blockchain since genesis. Thus, once it becomes able to participate in the network, it already has all the history stored locally.
+
+Fast Sync is going to change that picture. Once it starts being the method used for syncing, the joining nodes will only download some of the switch block headers and the global state from a recent block, ignoring a huge part of the network's history. This means that if a fast syncing node didn't have all the history already, it won't have at least some of it afterwards, either.
+
+Since Fast Sync has to be able to download full data of at least a single block, we propose for the archival nodes to use Fast Sync capabilities to sync block by block, from genesis up to the most recent blocks.
+
+The process will be analogous to the pre-fast-sync Linear Chain Sync:
+
+1. Walk back from trusted hash back to genesis.
+2. Download blocks one at a time starting from genesis upwards.
+    - The difference here would be that instead of executing the blocks, we would just download the global state trie from our peers.
+3. When we're in the current protocol version and can't fetch any more blocks, stop.
+
+After completing this process, the node will have the complete chain and all historical global state tries.
+
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+Going into more details of the process explained above, it would look like this:
+
+1. Download the header corresponding to the trusted hash.
+2. While the current header's height isn't 0, download the parent header by hash.
+3. When height 0 is reached, read the genesis set of validators for validating finality signatures.
+4. Download subsequent blocks with all the deploys, the global state trie and finality signatures in a loop.
+5. When no next block is available, stop.
+
+If we reach a block with a newer protocol version than our own during this process, we stop and attempt to upgrade.
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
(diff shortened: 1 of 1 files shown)


text/0066-archival-nodes.md
```diff
@@ -0,0 +1,86 @@
...
+With this design, there is a larger risk of losing the history of the network, as the number of complete copies would be dramatically decreased. However, the risk will still be small, and the lowered costs of running a regular nodes seem to be worth it.
+
+## Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+Archival nodes have to download _everything_ when syncing, so we need a process which satisfies that condition. The pre-fast-sync process achieved this goal, so an alternative could be to keep it and support it alongside fast sync, but:
+
+- The process detailed above is essentially equivalent to the old syncing process.
+- The old syncing process would have to be improved.
```
goral09, line 65: I may be biased since I'm the main author of the original linear chain sync but I still don't see what would have to be _improved_ . It's capable of synchronizing thousands of blocks in the matter of minutes without running out of memory, overloading peers, fairly distributing load across the network, disconnecting from invalid peers. All of that while providing a lot of metrics into the current state of the process (progress, memory footprint, etc.). It's resilient to restarts, failures, works across upgrades/multiple protocol versions, etc.

text/0066-archival-nodes.md
```diff
@@ -0,0 +1,86 @@
...
+- Keeping the two alongside each other is more code to be maintained.
```
goral09, line 66: That statement doesn't have strong arguments backing it up b/c we never made an effort to support it. It was gated behind a compilation flag which made the process much more difficult as APIs of "legacy" sync and fast sync had to match. If we made a decision to support both behind a _feature_ flag then it would have been much simpler.

text/0066-archival-nodes.md
```diff
@@ -0,0 +1,86 @@
...
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+Going into more details of the process explained above, it would look like this:
+
+1. Download the header corresponding to the trusted hash.
+2. While the current header's height isn't 0, download the parent header by hash.
+3. When height 0 is reached, read the genesis set of validators for validating finality signatures.
+4. Download subsequent blocks with all the deploys, the global state trie and finality signatures in a loop.
```
goral09, line 47: Why do we have to download deploys and finality signatures as well? What are the requirements for the archival node? Does it have to have all of the data:
* blocks
* deploys
* finality signatures
* events?

text/0066-archival-nodes.md
```diff
@@ -0,0 +1,86 @@
...
+5. When no next block is available, stop.
+
+If we reach a block with a newer protocol version than our own during this process, we stop and attempt to upgrade.
```
goral09, line 50: Why do we need to upgrade if we're not executing blocks? When we walk back the headers back to Genesis, do we downgrade as well?

text/0066-archival-nodes.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+Going into more details of the process explained above, it would look like this:
+
+1. Download the header corresponding to the trusted hash.
+2. While the current header's height isn't 0, download the parent header by hash.
+3. When height 0 is reached, read the genesis set of validators for validating finality signatures.
```
afck, line 46: We don't actually have to validate the finality signatures before the trusted block. (And signatures of unbonded validators shouldn't be trusted anymore anyway.)

text/0066-archival-nodes.md
fizyk20: More code is always more maintenance, and keeping the two alongside each other is more code, even if we switch between the two with a feature flag. We'll have to be constantly making sure that both systems still work, and that nodes using different systems within a single network don't conflict with each other (as we'll want to mix archival and non-archival nodes in one network).

text/0066-archival-nodes.md
fizyk20: Which events do you mean? I think the requirements are to keep at least all the blocks and deploys, and probably finality signatures, too.

text/0066-archival-nodes.md
fizyk20: Good point, but this part will get rewritten soon.

text/0066-archival-nodes.md
fizyk20: Per our call, this will get rewritten as a top-down sync.

afck (approved):

text/0066-archival-nodes.md
```diff
@@ -0,0 +1,85 @@
...
+    - Loop:
+        - download the block's global state trie, deploys and finality signatures, if not already present in storage
+        - if the current block height is 0, break
+        - set the current block to the current block's parent
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+With this design, there is a larger risk of losing the history of the network, as the number of complete copies would be dramatically decreased. However, the risk will still be small, and the lowered costs of running a regular nodes seem to be worth it.
```
KillianH, line 55: What solutions will be put in place to mitigate this risk ?

marc-casperlabs: I don't think this CEP in particular affects this risk - without the change, archival nodes would simply become operational later (after having downloaded all blocks, not directly after they have acquired enough state) instead. The core difference is that with fast sync, it is now _possible_ to run nodes with less history, so hopefully this will be a net increase, enabling those with smaller budgets to run nodes as well.

Arguably this change improves the situation for archival nodes as well, as it limits the cost of running an archival node by making it available sooner --- operators that might have been discouraged by the long startup time for an archival node but otherwise have no issue providing the necessary disk space now no longer need to make that tradeoff.

marc-casperlabs (approved):
Looks good to me at this point!

text/0066-archival-nodes.md
KillianH: I mean there won't be any advantage to run full archival nodes when this change will be up. So unless good Samaritan keeps some full archival nodes we will effectively lose some history ?

goral09: There's a number of products that do benefit from having the full history - data aggregators, block explorers, etc.. From the PoV of the protocol/validators, the old history is unnecessary.

KillianH: So does the association / partners are already planning to provide/keep full archival nodes ?

EdHastingsCasperLabs: Yes.

RitaMAllenCA:
@marc-casperlabs this is very old and appears to be done - may I close it?  Or you can...

RitaMAllenCA:
@fizyk20 can I close this?  It is really old and looks like it is documenting the standard?

Fraser999 (approved):
Approved to allow this to be merged.

PR #68, opened 2021-09-13
Delayed refunds

AlexanderLimonov:
A new method for issuing refunds, introducing dynamic penalties based on observed past refund percentages

### DIFF
--- a/text/delayed-refunds.md
+++ b/text/delayed-refunds.md
@@ -1,0 +1,104 @@
+# Delayed refunds
+
+## Summary
+
+[summary]: #summary
+
+CEP PR: [casperlabs/ceps# ](https://github.com/casperlabs/ceps/pull/)
+
+We propose enabling the dormant gas refund feature, additionally introducing a variable delay to disbursement of refunded tokens which depends on recently observed user behavior.
+
+## Motivation
+
+[motivation]: #motivation
+
+Due to the inherent uncertainty in gas costs caused by variation in inputs, users almost always need to overclaim gas to eliminate risk of reversion. Because refunds are always set to 0, this creates a serious UX problem. Additionally, a complete resolution to this problem would require full refunds. Unfortunately, given the execution-after-consensus paradigm Casper uses, full refunds create a serious attack vector, enabling malicious or careless users to fill up blocks with overclaimed gas.
+
+To address both the UX issue, and the security issue with refunds, we propose a delay to disbursement of refunded tokens, which would always be held for at least one era. However, depending on the proportion of gas refunded to the user, the delay would grow longer. The length of the delay is governed by the amount of gas recently refunded.
+
+## Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+### System view
+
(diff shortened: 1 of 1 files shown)


text/0068-delayed-refunds.md
```diff
@@ -0,0 +1,104 @@
...
+
+[reference-level-explanation]: #reference-level-explanation
+
+### Refund delay logic
+
+We will denote state variables evolving as eras switch in bold letters.
+
+#### Elements
+
+*Thresholds* in {p<sub>1</sub>, ..., p<sub>N</sub>} are percentages (fractions ranging from 0 to 1) that define ranges of observed past refunds, which determine a *penalty* in {1, ..., M}.
```
afck, line 41: > ranges of observed past refunds

I'm not sure I understand correctly; maybe this could be clarified in the text:

The p<sub>i</sub> are fixed parameters, are they? They are not the actual observed refunds?

E.g. we could configure (in the chainspec) {p<sub>1</sub>, ..., p<sub>N</sub>} to be {10%, 20%, 50%}, so that penalty 1 applies if a user reserved 10% too much, 2 if it's 20% too much and 3 if it's 50% too much?

text/0068-delayed-refunds.md
```diff
@@ -0,0 +1,104 @@
...
+
+Fix N to be #{1, ..., M}. The thresholds define the penalty in the following way:
```
afck, line 43: So… N = M?

text/0068-delayed-refunds.md
```diff
@@ -0,0 +1,104 @@
...
+
+- Take p<sub>i</sub> and p<sub>i+1</sub>
+- If one minus *aggregate utilization* **P** (1 - **P**) is between p<sub>i</sub> and p<sub>i+1</sub>
```
afck, line 46: I read this as "P · (1 - P)", i.e. "P times one minus P" at first. Maybe less ambiguous:

```suggestion
- If 1 - **P**, i.e. one minus *aggregate utilization* **P**, is between p<sub>i</sub> and p<sub>i+1</sub>
```

text/0068-delayed-refunds.md
AlexanderLimonov: Yes, it's just a step function laid out in the chainspec, exactly as you describe, will clarify.

text/0068-delayed-refunds.md
AlexanderLimonov: Awkwardness due to M not necessarily being an integer,

text/0068-delayed-refunds.md
AlexanderLimonov: Agreed

text/0068-delayed-refunds.md
afck: But then why does it start with 1? Does that mean "a penalty of 1 era", i.e. delaying the refund by one era?

AlexanderLimonov: The penalty is multiplicative, so 1 means no delay at all (everything is refunded at the end of the era). I do see where the problem is now, though, since non-integer multipliers on number of eras do not make sense.

RitaMAllenCA:
@AlexanderLimonov can this be closed?  Merged?

AlexanderLimonov:
There's no reason to touch it either way right now. This work will resume after finality signature rewards are done.

MParlikar (changes requested):
@AlexanderLimonov this needs to be updated to reflect updated thinking.  Please also include the updates to RPC and the precise refund mechanism.

PR #74, opened 2021-11-16
first draft of transfer memo CEP.

nilz3ro:


text/0074-alphanumeric-transfer-ids.md
```diff
@@ -0,0 +1,130 @@
...
+
+[drawbacks]: #drawbacks
+
+Major changes to core features of the blockchain network require significant time and effort to implement, and may also come with unforeseen consequences.
+
+## Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The `Memo` type can be implemented without requiring any data migrations because of the way that the current `Option` tag bytes correspond to the new `Memo` tag bytes. Adding a new, independent text field to `Transaction` would require a production data migration, and it could be confusing to users that are trying to understand the difference between the optional integral type `id` field and the alphanumeric memo field.
```
goral09, line 112: Personally, I think that the `Memo` field should be added to a _deploy_ type and not `Transfer` variant. All transers are deploys but not all deploys are transfers. By adding the `Memo` to transfer only we're limiting users' capabilities. It's quite a common practice to attach memo to a contract call on other networks. For example, in Terra it's [part of transaction body](https://github.com/terra-money/terra.js/blob/9e5f553de3ff3e975eaaf91b1f06e45658b1a5e0/src/client/lcd/api/TxAPI.ts#L247).

text/0074-alphanumeric-transfer-ids.md
```diff
@@ -0,0 +1,130 @@
...
+}
+```
+
+`Transfer` would then be changed to hold a non-optional `Memo` instead of the current `Option<u64>`. This relies on the way that we're currently serializing `Option<64>`. The tag for `Option::None` corresponds to `Memo::None` and the `Option::Some` tag corresponds to `Memo::Numeric`. The tag for `Memo::Alphanumeric` is new and will not conflict with the existing tags.
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+Major changes to core features of the blockchain network require significant time and effort to implement, and may also come with unforeseen consequences.
```
goral09, line 106: This will require changing all SDKs to accept new types and in some cases the APIs might not be backwards compatible - forcing all users to upgrade.

text/0074-alphanumeric-transfer-ids.md
```diff
@@ -0,0 +1,130 @@
...
+Here is an example implementation for a new `Memo` type that will be stored in a `Transfer`'s `id` field.
+
+```rust
+/// Memo field associated with a [`Transfer`].
+#[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Serialize, Deserialize)]
+#[cfg_attr(feature = "json-schema", derive(JsonSchema))]
+pub enum Memo {
+    None,
+    Numeric(u64),
+    Alphanumeric(String),
```
goral09, line 38: Will there be a limit on the size of this? Or is it going to be unbounded? What encoding will it use, utf-8?

piotr-dziubecki:
@EdHastingsCasperLabs we will rework this concept in 2.0

PR #77, opened 2022-04-04
CEP77 - Make eras longer

KillianH:
[Rendered](https://github.com/KillianH/ceps/blob/longer-eras/text/0077-longer-eras.md)


sacherjj:
Drawbacks is slower reaction to bad nodes which drastically increases the risk of holding nodes which should be ejected.

PR #82, opened 2023-02-17
WIP: CEP-82: Trading Extension to CEP-78

ElusiveMori:
[rendered](https://github.com/ElusiveMori/ceps/blob/trading/text/0082-cep-78-trading-extension.md)


text/0082-cep-78-trading-extension.md
```diff
@@ -0,0 +1,115 @@
...
+* `CustodyTransfer` - custody via transfer
+
+Taking ownership of the NFT is needed in cases where it is necessary to prevent the owner of the NFT from unexpectedly transferring the NFT to a third party, such as during an auction. Otherwise, approvals are more appropriate, as they do not change the actual owner of the NFT. 
+
+_(wip) remark: this requirement might conflict with the potential implementation of forced royalty collection using an intermediary custodial/royalties contract_
+
+### Duration
+
+Another dimension that can differentiate `Marketplace` contracts is _duration_:
+* A `Marketplace` can be _instant_, meaning that an exchange of items can happen as soon as two parties agree, such as in a swap. A consequence of this is that only the _seller_ needs to hand over custody of the item being sold, whereas the _buyer_ operation can be atomic. `CustodyApproval` will suffice for this `Marketplace`
```
sfeast, line 43: If I'm reading this correctly, then with only `CustodyApproval` I think there is potential for the order book to become out of sync if the NFT owner directly transfers the item while it is listed, which has several ramifications depending on the order book implementation.

ElusiveMori: This is true. To be fair, I'm struggling to find a use case where using the `approve` mechanism would make sense.

Though in a very simplistic implementation, the offending item could simply be removed upon trying to purchase it.

sfeast: Yes but doesn't lead to the best market UX - user tries to buy, the transaction fails, they get charged a fee & find out the item isn't actually for sale 😓  

Also for 3rd party dapps & services they would not have access to a correct orderbook while items are incorrectly listed. Assuming there will be an orderbook index to access.

text/0082-cep-78-trading-extension.md
```diff
@@ -0,0 +1,115 @@
...
+* A `Marketplace` can be _continuous_ (find a better word?), meaning that an exchange can happen only when a duration of time has elapsed or multiple parties have participated, such as in an auction. In general, this will require that each party give the marketplace full custody of the asset being traded, as trying to handle the edge cases where an asset disappears from a running auction would introduce unnecessary complexity. `CustodyTransfer` is required for this `Marketplace`.
+
+Similarly, this introduces two distinct modalities for a `Marketplace`: 
+* `DurationInstant`
+* `DurationContinuous`
+
+### Interface / Key Functions
```
darthsiroftardis, line 50: We need to clearly disambiguate these entry point names because they *will* collide with the entry points on the systems auction contract, something like "post_marketplace", "bid_on_nft"

RitaMAllenCA:
Hi - please close this PR if the standard is complete and approved.  Thanks

MParlikar:
@gRoussac please take a look

PR #86, opened 2023-08-14
Factory Pattern

mpapierski:
See also casper-network/casper-node#4181

[Rendered](https://github.com/mpapierski/ceps/blob/factory-pattern/text/0086-factory-pattern.md)

### DIFF
--- a/text/0086-factory-pattern.md
+++ b/text/0086-factory-pattern.md
@@ -1,0 +1,87 @@
+# Factory Pattern
+
+## Summary
+
+[summary]: #summary
+
+CEP PR: [casper-network/ceps#0086](https://github.com/casper-network/ceps/pull/86)
+
+One paragraph explanation of the feature.
+
+## Motivation
+
+[motivation]: #motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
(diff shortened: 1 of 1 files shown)


text/0086-factory-pattern.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+In technical detail, this proposal introduces two new entry point types: `EntryPointType::Install` and `EntryPointType::Normal`. And a new entry point access type `EntryPointAccess::Abstract`.
+
+* `EntryPointType::Install` marks an entry point as a factory method responsible for deploying contracts.
+* `EntryPointType::Normal` behaves similarly to `EntryPointType::Contract` in version 1.x, while also serving as the default entry point type in version 2.0.
```
darthsiroftardis, line 34: Are we replacing the `Contract` variant with `Normal`, I think we should be specific here in saying what the go-forward will be.

darthsiroftardis (approved):

text/0086-factory-pattern.md
mpapierski: Done

PR #87, opened 2023-09-21
Add data types extensibility CEP

rafal-ch:
CEP for Data Extensibility.

Closes https://github.com/casper-network/casper-node/issues/2941

[Rednered](https://github.com/casper-network/ceps/blob/66bbb21ab776d73eb17069cb86fb497092317992/text/0087-data-types-extensibility.md)

### DIFF
--- a/text/0087-data-types-extensibility.md
+++ b/text/0087-data-types-extensibility.md
@@ -72,7 +72,7 @@ Components that never need to touch the legacy types should directly use the mos
 
 The `Fetcher` on the other hand must be able to deal with all variants of the datatype, because it must be able to fetch and understand all historical blocks up to genesis.
 
-## Storage
+### Storage
 
 Storage is the main component that should understand and work with the versioned types. For each data type that needs to be extensible we create a separate DB which will be used to store the top-level `enum` wrapper. The legacy DB is left intact and it's going to be accessed only when old-enough data is requested. Therefore, upon the request to the storage being sent, storage will first look for a particular item in the new database and only if it is not found it'll reach out to the legacy DB and try to obtain data from there.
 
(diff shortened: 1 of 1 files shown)


text/0087-data-types-extensibility.md
```diff
@@ -0,0 +1,130 @@
...
+        }
+        Ok(buffer)
+    }
+```
+
+Components that never need to touch the legacy types should directly use the most recent version of the datatype. For example, going forward, we never need to gossip "legacy" blocks during the normal chain operation, therefore the `Gossiper` should work directly with `BlockV2`.
+
+The `Fetcher` on the other hand must be able to deal with all variants of the datatype, because it must be able to fetch and understand all historical blocks up to genesis.
+
+## Storage
```
alsrdn, line 75: nit: I think this should have heading 3 `###`

text/0087-data-types-extensibility.md
```diff
@@ -0,0 +1,130 @@
...
+    else {
+        body
+    }
+}
+```
+There is no noticeable performance degradation expected because most of the requests will be saturated from the new DB and we'll fall back to the legacy one only for limited use cases (like historical sync, which in most cases is a one-time operation).
+
+Upon storage initialization, indices are built based on the content of both DBs. Similarly, all utility functions, like `value_exists()` will be updated to check both databases.
+
+# Alternatives
```
alsrdn, line 96: Since the title of the CEP has heading 1 I think that this section should have heading 2 `##` and all of its subsections should have heading 3

alsrdn (approved):
A couple of formatting issues but I think the contents are sound

text/0087-data-types-extensibility.md
rafal-ch: Headings fixed in https://github.com/casper-network/ceps/pull/87/commits/a1e164f0a5f7c0ad1a170f6f112ba5759b6eb6de

text/0087-data-types-extensibility.md
rafal-ch: Headings fixed in https://github.com/casper-network/ceps/pull/87/commits/a1e164f0a5f7c0ad1a170f6f112ba5759b6eb6de

Fraser999 (approved):
Approved to allow this to be merged.

PR #88, opened 2023-09-25
Add a draft for the contract level messages CEP.

alsrdn:
This PR adds a draft for introducing contract level messages whcih are a facility that enable a smart contract to emit a message while executing which eventually gets sent on the event stream and is visible from outside the network.

Rendered: https://github.com/casper-network/ceps/blob/943cbf511e5ae3bc195cb43fcc4cea313d5ea3a1/text/0088-contract-level-messages.md


text/0088-contract-level-messages.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+The `Message` enum initially allows for sending human readable strings as messages in the first iteration:
+```rust
+pub enum Message {
+    String(String)
+}
+```
+
+#### A new `Key` variant is introduced for indexing the message topic digests:
```
xcthulhu, line 84: First, thank you for the excellent CEP.

I would suggest a tweak to the architecture. Rather than include execution effects in global storage, I would recommend we construct a separate Merkle tree of deploy execution effects and include a Merkle root in each block (following Ethereum's architecture for transaction receipts). We would not have to persist this Merkle tree, we would compute the Merkle root for it and only persist canonical deploy execution effects (indexed by deploy) in storage.

Does that make sense?

alsrdn: Thank you for the feedback! Your proposal is interesting.
However going through some experiments it became apparent that we would probably want to allow users to query for checksums of distinct messages.
With the previous design the application listening for messages on a topic would have needed to catch all messages to be able to verify the whole topic which is a bit unrealistic.

So with this new requirement I think we would have to also persist this new Merkle tree that you are proposing as well to be able to query for any discrete message checksum. Not sure if there's any gain in having it separate rather than just writing to global state in this case.

I've done some modifications to this CEP in the latest commit so please take a look; the changes are mainly related to how we validate these discrete messages and added a bit more info on gas cost and limits.

text/0088-contract-level-messages.md
```diff
@@ -0,0 +1,237 @@
...
+Checksum of the message payload that was emitted:
+```rust
+pub struct MessageChecksum(
+    pub  [u8; 32], //a 32 byte BLAKE2b hash of the serialized `MessagePayload` that was emitted.
+);
+```
+
+When a new message is emitted on a specified topic the `message_count` in the `MessageTopicSummary` record is incremented.
+Messages have ephemeral nature by design. In order to not increase the size of global state in an uncontrolled manner, the message count is reset to zero with every newly emitted message that belongs to a new block and the block time is updated as well. The old message checksum entries are pruned from global state.
+Of-chain applications can still query the old message checksums if required by using the state root hash of the block where the messages were emitted.
```
Fraser999, line 139:
```suggestion
Off-chain applications can still query the old message checksums if required by using the state root hash of the block where the messages were emitted.
```

text/0088-contract-level-messages.md
```diff
@@ -0,0 +1,237 @@
...
+
+The `TransactionProcessed` SSE event will contain a `messages` field that will have an array of `Message` objects:
+
+```rust
+pub struct Message {
+    /// The identity of the entity that produced the message.
+    entity_addr: AddressableEntityHash,
+    /// The payload of the message.
+    message: MessagePayload,
+    /// The name of the topic on which the message was emitted on.
```
Fraser999, line 155:
```suggestion
    /// The name of the topic on which the message was emitted.
```

Fraser999 (approved):
Nice job - very concise explanation.

text/0088-contract-level-messages.md
alsrdn: Fixed in: https://github.com/casper-network/ceps/pull/88/commits/5ed76edf97bdf5e31cada98ce01522dad3e109d5

text/0088-contract-level-messages.md
alsrdn: Fixed in: https://github.com/casper-network/ceps/pull/88/commits/5ed76edf97bdf5e31cada98ce01522dad3e109d5

rafal-ch (approved):
LGTM :+1:

text/0088-contract-level-messages.md
xcthulhu: > So with this new requirement I think we would have to also persist this new Merkle tree that you are proposing as well to be able to query for any discrete message checksum. Not sure if there's any gain in having it separate rather than just writing to global state in this case.

You would not need to persist this Merkle tree -- Ethereum does not persist Merkle trees for its contract level events.  You just compute the Merkle root and store the events.  There is already an implementation of this in the `casper-node` repo: https://github.com/casper-network/casper-node/blob/29466f9089d309720e2d8515a3b649a1e0a4fbae/hashing/src/lib.rs#L156

In addition, have you looked at [Ethereum's bloom filter design](https://ethereum.github.io/execution-specs/autoapi/ethereum/frontier/bloom/index.html) for its events?  This is commonly used by light clients as it avoids the need to download all contract logs when checking a block.

xcthulhu: There's also code for computing a Merkle proof of inclusion given a Merkle root: https://github.com/casper-network/casper-node/blob/29466f9089d309720e2d8515a3b649a1e0a4fbae/hashing/src/indexed_merkle_proof.rs#L63

alsrdn: > > So with this new requirement I think we would have to also persist this new Merkle tree that you are proposing as well to be able to query for any discrete message checksum. Not sure if there's any gain in having it separate rather than just writing to global state in this case.
> 
> You would not need to persist this Merkle tree -- Ethereum does not persist Merkle trees for its contract level events. You just compute the Merkle root and store the events. There is already an implementation of this in the `casper-node` repo: https://github.com/casper-network/casper-node/blob/29466f9089d309720e2d8515a3b649a1e0a4fbae/hashing/src/lib.rs#L156

We wouldn't want to store the full events though. For this to provide an advantage we would probably want to pass on the merkle proof along with the events on the SSE to allow consumers to verify it. But most likely we would want to store some information (e.g. the number of messages that were emitted on each topic in order to see if a particular type of message was emitted at all). Given this I think we get some additional advantage if we store this type of info in Global State as proposed in the CEP: since consumers can listen for the execution journal, they can more accurately reconstruct the context in which that message was emitted should they require to.
This gives a bit more transparency to the system rather than providing only proofs for the events emitted that are rather opaque WRT context.

text/0088-contract-level-messages.md
```diff
@@ -0,0 +1,237 @@
...
+#### Introduce new FFIs:
+```rust
+/// Manages a message topic with the specified name for the calling contract.
+/// 
+/// The topic name is a string of a maximum length that is specified in the chainspec.
+/// The initially supported operation will be `Add` that registers a new message topic.
+/// Will return an error if:
+///     * the topic name length is exceeded
+///     * the topic already exists (in case of the Add operation)
+///     * the call was made from an `Account` context. We only allow stored contracts to register messages.
```
zie1ony, line 62: I'd remove this constraint. It would be useful to emit new event with the new contract address when installing new contract in `call`, which is in the `Account` context.

text/0088-contract-level-messages.md
```diff
@@ -0,0 +1,237 @@
...
+    message_ptr: *const u8,
+    message_size: usize,
+);
+```
+
+The `MessagePayload` enum initially allows for sending human readable strings as messages in the first iteration:
+```rust
+pub enum MessagePayload {
+    /// Human readable string message.
+    String(String),
```
zie1ony, line 91:
```suggestion
    Bytes(casper_types::bytesrepr::Bytes),
```

Events are blobs of bytes (think CEP18 events). Having them as `String` will force me to to hex or base64 bytes before emitting. It just adds unnecessary gas.

text/0088-contract-level-messages.md
```diff
@@ -0,0 +1,237 @@
...
+```rust
+/// Manages a message topic with the specified name for the calling contract.
+/// 
+/// The topic name is a string of a maximum length that is specified in the chainspec.
+/// The initially supported operation will be `Add` that registers a new message topic.
+/// Will return an error if:
+///     * the topic name length is exceeded
+///     * the topic already exists (in case of the Add operation)
+///     * the call was made from an `Account` context. We only allow stored contracts to register messages.
+pub fn casper_manage_message_topic(
```
zie1ony, line 63: I'd remove this FFI in favour of registering all events at the contract installation step along side the entry points information. It will enable consumer to know what kind of event-data-schemas (at least names) they need before they start consuming events. Most of the time (all?) you know all the events your contract will emit at the compilation point. I'd even say that making on-the-fly topics is rather undesirable.

text/0088-contract-level-messages.md
```diff
@@ -0,0 +1,237 @@
...
+pub struct Message {
+    /// The identity of the entity that produced the message.
+    entity_addr: AddressableEntityHash,
+    /// The payload of the message.
+    message: MessagePayload,
+    /// The name of the topic on which the message was emitted.
+    topic_name: String,
+    /// The hash of the name of the topic.
+    topic_name_hash: TopicNameHash,
+    /// Message index in the topic.
```
zie1ony, line 159: `Message index in the topic` is not very useful. What is useful is having the global index for all events emitted by the contract, so I can reason about a causality.

Just to be clear I'd crate separate topics for every event type.

zie1ony (changes requested):
Great progress 🔥
From the smart contract developer point of view it is better then what we have now, but interfaces can be more useful.

text/0088-contract-level-messages.md
alsrdn: I had discussed this with @EdHastingsCasperAssociation in the past as well. The conclusion was that maybe it would be more helpful in the future iteration (and as an extension to the contract level messages functionality) to have the host emit standard system level messages when a certain type of change is performed (e.g. account is created, contract is installed etc).

text/0088-contract-level-messages.md
alsrdn: Yes, it's probably a good idea to add the `Bytes` variant alongside `String` from the start. The initial requirement was to be able to emit human readable strings, we made the payload an enum in order to allow easy expansion in the future. The enhancement would probably also include standard payloads for system type messages described above.

text/0088-contract-level-messages.md
alsrdn: Yes, we could do it but it would maybe reduce some of the flexibility. I'm guessing some contracts would rather delay registering topics until actually needed (because registering topics also cost gas). Also we would still want to have some kind of FFI for managing message topics in case we want to add some more operations in the future.

text/0088-contract-level-messages.md
alsrdn: We would still need the message index in the topic to be able to query the checksum. But you make a good point about cross-topic correlation of messages.
I'm thinking maybe of a use-case where you want to check that a particular "burn" message for a token id was emitted only after a "mint" message for the same id?
Having a global counter per contract would solve the sequence correlation out of the box, but at the same time it has the potential to bring some noise if a large number of unrelated topics are used.

With the current proposal you could have a separate 3rd topic used to log the sequence of the messages in the other 2 topics which should also solve the issue and have a minimal amount of noise.

text/0088-contract-level-messages.md
zie1ony: System level events that would cover all the actions user-actions is a good idea.

text/0088-contract-level-messages.md
zie1ony: Thanks. Operations on strings in smart contracts are just waste of gas.

text/0088-contract-level-messages.md
zie1ony: I'd introduce new index: `all_topics_index` so both are possible. `u32` is large enough for counting all events.

If the solution is having 3rd topic, I'd probably end up having just one contract-wide topic and push all the events into it. And then the whole topic feature doesn't really make sense.

Another uncovered thing is the correlation of events between subcalls. I'd like to know what triggered what (like in etherscan presents).

Consider this scenario:
```
1. Contract A emits EventA(data1).
2. Contract A calls Contract B.
3. Contract B emits EventB.
4. Contract B calls Contract (A -> B -> A)
5. Contract A emits EventA(data2).
```

I'd like to be able to easily know the sequence:
```
1. ContractA, EventA(data1)
2. ContractB, EventB
3. ContractA, EventA(data2)
```

text/0088-contract-level-messages.md
zie1ony: > Why would registering topics on the fly be undesirable?

1. If that would be lazy, then emitting first event requires the topic registration and this additional gas. As someone who calls contract I want the gas costs to predictable and constant as much as possible.
2. If that would be lazy, then how do I know if the topic was already registered? You'd have to expose new `ffi` that checks if topic is already registered. Then before every call I'd have to make this check, so every event emit is two `ffi` calls.
3. Because of the above point nr 2, I'd have to register all the topics as soon as possible. Casper doesn't have constructors, but we very often mimics them by calling a `constructor` entrypoint just after the contract installation in `call` and then revoking the access to `constructor`, so it can't be called again. Not every contract needs calling a constructor (not calling it saves gas), but now I'd have to account for events registration and call the `constructor` also when contract needs topics registration.

text/0088-contract-level-messages.md
alsrdn: I've mentioned the system level events as a future work item in https://github.com/casper-network/ceps/pull/88/commits/afaf653819ca6db9bf68415586ffd93aec2dd8a8

text/0088-contract-level-messages.md
alsrdn: Added the variant to hold `Bytes` for the payload: https://github.com/casper-network/ceps/pull/88/commits/afaf653819ca6db9bf68415586ffd93aec2dd8a8

text/0088-contract-level-messages.md
alsrdn: Changed the FFI to allow messages to be managed on contract install/upgrade as per request: https://github.com/casper-network/ceps/pull/88/commits/afaf653819ca6db9bf68415586ffd93aec2dd8a8

text/0088-contract-level-messages.md
alsrdn: I added a section to the CEP related to tracking messages across contracts and topics. Please take a look here: https://github.com/casper-network/ceps/pull/88/commits/afaf653819ca6db9bf68415586ffd93aec2dd8a8

The proposal I wrote involves writing a log for the message keys that is indexed by the deploy hash. This would allow consumers to enable this functionality and track messages across contract invocations.
Please take a look.

text/0088-contract-level-messages.md
```diff
@@ -0,0 +1,306 @@
...
+Consumers can query global state by transaction hash and message sequence to get the keys of the individual messages that were emitted by the contracts. A summary value will also be available for consumers to be able to determine how many messages were emitted as part of the transaction.
+
+If for example a transaction is processed where two message emitting contracts will be called, the following entries will be written to global state:
+  * `Key::Message(contract_A, topic_A_hash, Some(0))` -> `StoredValue::Message(message_A_0_checksum)`
+  * `Key::Message(contract_A, topic_A_hash, None)` -> `StoredValue::MessageTopic(MessageTopicSummary { message_count: 1, block_time: 12345 })`
+  * `Key::Message(contract_B, topic_B_hash, Some(0))` -> `StoredValue::Message(message_B_0_checksum)`
+  * `Key::Message(contract_B, topic_B_hash, None)` -> `StoredValue::MessageTopic(MessageTopicSummary { message_count: 1, block_time: 12345 })`
+  * `Key::Message(contract_B, topic_B_hash, Some(1))` -> `StoredValue::Message(message_1_0_checksum)`
```
rafal-ch, line 270:
```suggestion
  * `Key::Message(contract_B, topic_B_hash, Some(1))` -> `StoredValue::Message(message_B_1_checksum)`
```
?

text/0088-contract-level-messages.md
```diff
@@ -0,0 +1,306 @@
...
+  * `Key::Message(contract_B, topic_B_hash, None)` -> `StoredValue::MessageTopic(MessageTopicSummary { message_count: 1, block_time: 12345 })`
```
rafal-ch, line 271:
```suggestion
  * `Key::Message(contract_B, topic_B_hash, None)` -> `StoredValue::MessageTopic(MessageTopicSummary { message_count: 2, block_time: 12345 })`
```
?

text/0088-contract-level-messages.md
alsrdn: This was removed in favour of the changes included here: https://github.com/casper-network/ceps/pull/88/commits/a5f40e7155ecab5995d9c5ea1e57027e5ded8545

text/0088-contract-level-messages.md
alsrdn: This was removed in favour of the changes included here: https://github.com/casper-network/ceps/pull/88/commits/a5f40e7155ecab5995d9c5ea1e57027e5ded8545

text/0088-contract-level-messages.md
```diff
@@ -0,0 +1,312 @@
...
+    ...
+    /// A `Key` under which the total number of emitted messages in the last block is stored.
+    BlockMessageCount,
+}
+```
+
+Each message that is emitted within a single block has an unique sequence number. This record is reset with every new created block.
+The value written under this key is a tuple type `CLValue` representing the `(block_time, block_message_count)`.
+The `block_message_count`represents the total number of messages emitted in the block with the corresponding `block_time`.
+The sequence id of the messages are global to the block and are not tied to entity address or topic.
```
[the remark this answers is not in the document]
alsrdn: fixed typo in: https://github.com/casper-network/ceps/pull/88/commits/6b6e0735ac40dc9b70f2320d25a45b66bc154126

PR #93, opened 2024-03-06
Sync CEP18 proposal with current implementation

gyroflaw:
## Background

It is important to have a truth source for every CEP. While was checking CEP18 implementation, I found that it is quite different from original proposal.

## Changes
- Removed batch entry points.
- Added `increase_allownace` and `decrease_allowance` entry points.


ACStoneCL (approved):
Makes sense to me.

ipopescu (approved):

PR #95, opened 2025-04-14
Initial proposal for CEP-95 NFT token standard

Volodymyr-Kuchinskyi:
[Rendered](https://github.com/make-software/casper-ceps/blob/CEP-93-standard/text/0095-nft-standard.md)

### DIFF
--- a/text/0095-nft-standard.md
+++ b/text/0095-nft-standard.md
@@ -29,7 +29,8 @@ The objective of this proposal is to define a Casper-compatible NFT standard tha
 - Tracking ownership of NFTs.
 - Transferring NFTs safely between accounts and contracts.
 - Managing per-token approvals as well as operator-level approvals.
-- (Optionally) exposing metadata (name, symbol, token URI) that describes the underlying asset details.
+- Exposing token contract metadata (token_name, token_symbol).
+- (Optionally) exposing token metadata that describes the underlying asset details (name, description, asset_uri, token_uri).
 
 This proposal takes into account a possibility of off-chain indexing of NFTs by specifying interfaces of the events that can be used to track the contract activity.
 ## Specification
@@ -69,7 +70,10 @@ pub enum CEP95Event {
 
 /// Casper-compatible NFT interface
 pub trait CEP95 {
-    /// Returns an optional short symbol or abbreviation for the NFT collection.
+    /// Returns a name of the NFT token/collection.
+    fn name(&self) -> String;
+
+    /// Returns a short symbol or abbreviation for the NFT token/collection.
     fn symbol(&self) -> String;
 
     /// Returns the number of NFTs owned by a given account or contract
@@ -159,7 +163,7 @@ pub trait CEP95Receiver {
 }
 ```
 
-### Contract symbol metadata
+### Token Contract metadata
 
 To keep contract metadata storage consistent it is strongly recommended to follow the CEP-96 standard.
 
@@ -167,23 +171,35 @@ To keep contract metadata storage consistent it is strongly recommended to follo
 
 1. Install/Constructor Phase:
     ```rust
-    let symbol: String = runtime::get_named_arg("symbol");
+    let token_name: String = runtime::get_named_arg("name");
+    let token_symbol: String = runtime::get_named_arg("symbol");
     
-    let symbol_uref = storage::new_uref(symbol);
-    runtime::put_key("symbol", symbol_uref.into());
+    let token_name_uref = storage::new_uref(token_name);
+    runtime::put_key("name", token_name_uref.into());
+   
+    let token_symbol_uref = storage::new_uref(token_symbol);
+    runtime::put_key("symbol", token_symbol_uref.into());
     ```
 2. Interface Method:
     ```rust
-    #[no_mangle]
+    pub extern "C" fn name() {
+        let key = runtime::get_key("name").expect("name key not found");
+
+        let token_name_uref = key.into_uref().expect("name key is not a URef");
+        let token_name: String = storage::read(token_name_uref)
+            .unwrap_or_revert();       // read() -> Result<String, _>
+    
+        runtime::ret(CLValue::from_t(token_name).unwrap_or_revert());
+    }
+   
     pub extern "C" fn symbol() {
         let key = runtime::get_key("symbol").expect("symbol key not found");
-
-        let symbol_uref = key.into_uref().expect("symbol key is not a URef");
-        let symbol: String = storage::read(symbol_uref)
-            .unwrap_or_revert()       // read() -> Result<Option<T>, _>
-            .unwrap_or_revert();      // Option<T> -> T (revert if None)
     
-        runtime::ret(CLValue::from_t(symbol).unwrap_or_revert());
+        let token_symbol_uref = key.into_uref().expect("symbol key is not a URef");
+        let token_symbol: String = storage::read(token_symbol_uref)
+            .unwrap_or_revert();       // read() -> Result<String, _>
+    
+        runtime::ret(CLValue::from_t(token_symbol).unwrap_or_revert());
     }
     ```
 
@@ -258,6 +274,9 @@ pub trait CEP95TokenMetadata {
 }
 ```
 
+* Named Key for dictionary: `token_metadata`
+* Key Structure: token_metadata[<token_id>]
+
 ### Off-chain Token Metadata Schema
 
 ```json
@@ -277,6 +296,9 @@ pub trait CEP95TokenMetadata {
     "required": []
 }
 ```
+
+* Named Key for dictionary: `token_metadata`
+* Key Structure: token_metadata[<token_id>]
 
 ## Event Structure
 
@@ -350,6 +372,8 @@ pub enum CEP95MetadataEvent {
 
 Backend indexers or off-chain processors should listen for NFT-related events (`CEP95Event` event enum):
 
+Note: Important mention that conventionally CES (Casper Event Standard) events should be used by default, but in the future the spec might be updated adding native events.
+
 - **Minting**:
   - Detect a `Mint` event.
   - Extract the `token_id` and `to` address.
@@ -393,3 +417,4 @@ The Casper NFT Standard (inspired by ERC-721) aims to bring uniformity and inter
 - [CEP-96 Contract Metadata Standard](https://github.com/casper-network/ceps/pull/96)
 - [Casper Documentation (for contract APIs and types)](https://docs.casper.network/concepts/smart-contracts)
 - [Example proposals and CEPs from the Casper community](https://github.com/make-software/casper-ceps/tree/master/text)
+- [Casper Event Standard](https://github.com/make-software/casper-event-standard)
--- a/text/0095-nft-standard.md
+++ b/text/0095-nft-standard.md
@@ -4,7 +4,7 @@ Casper NFT Standard Proposal (Inspired by ERC-721 & adapted for the Casper ecosy
 
 ## Summary
 
-CEP PR: [casperlabs/ceps#0095]
+CEP PR: [casper-network/ceps#0095](https://github.com/casper-network/ceps/pull/95)
 
 This standard provides a non-fungible token (NFT) contract API that is consistent with the existing [CEP-18](https://github.com/casper-network/ceps/blob/master/text/0018-token-standard.md) fungible token standard. It aims to replace [CEP-47](https://github.com/casper-network/ceps/blob/master/text/0047-casper-nft-protocol.md) and [CEP-78](https://github.com/casper-network/ceps/blob/master/text/0078-enhanced-nft-standard.md), which have flaws that complicate their support in the ecosystem. CEP-47 lacks operator approval functionality, which makes it less convenient to integrate with NFT marketplaces. CEP-78 rather describes a configurable NFT contract that can modify its behavior to various needs, rather than a plain NFT contract API.
 
@@ -36,7 +36,7 @@ This proposal takes into account a possibility of off-chain indexing of NFTs by 
 
 [specification]: #specification
 
-### Core Casper NFT Interface
+### NFT Contract Interface
 
 [core-casper-nft-interface]: #core-casper-nft-interface
 
@@ -113,7 +113,7 @@ pub trait CEP95 {
     ///
     /// @param token_id - The NFT ID.
     /// @return Option<Key> - Approved spender account if one exists.
-    fn get_approved(&self, token_id: U256) -> Option<Key>;
+    fn approved_for(&self, token_id: U256) -> Option<Key>;
 
     /// Enables operator approval for all of the caller's NFTs.
     ///
@@ -134,7 +134,7 @@ pub trait CEP95 {
 }
 ```
 
-### Recommended Adoption of the Casper On-chain Contract Metadata Standard CEP-96
+### Contract Metadata Standard CEP-96
 
 To improve the integrity, discoverability, and user experience within the Casper ecosystem, implementers of the CEP-95 NFT Standard are strongly encouraged to also adopt the **CEP-96 Casper On-chain Contract Metadata Standard**.
 
@@ -142,9 +142,9 @@ By implementing these metadata standard, CEP-95 contracts become easily identifi
 
 We strongly recommend incorporating this metadata standard alongside CEP-95 implementations to contribute positively to ecosystem consistency, transparency, and user-friendliness.
 
-### Recommended Storage for contract data (symbol, etc.)
+### Contract symbol metadata
 
-According to the CEP-96 standard, it is recommended to store contract metadata as a form of contract level named keys using `Urefs`.
+To keep contract metadata storage consistent it is strongly recommended to follow the CEP-96 standard.
 
 **Example Named Key Usage (High-Level)**
 
@@ -183,9 +183,11 @@ According to the CEP-96 standard, it is recommended to store contract metadata a
 The CEP-95 standard provides a flexible, hybrid approach for NFT token metadata, allowing developers and consumers to leverage **both on-chain and off-chain metadata sources**.
 
 - **On-chain metadata** (`token_metadata`):
-  - This method allows direct storage of metadata within the blockchain. Metadata is stored as key-value pairs (`Vec<(String, String)>`), giving a clear yet flexible structure without enforcing a specific serialization format. When metadata is created or updated, the contract emits a `MetadataUpdate` event, ensuring that indexers and clients are immediately informed and can cache or display this data without additional lookups. 
+  - This method allows direct storage of metadata within the blockchain. Metadata is stored as key-value pairs (`Vec<(String, String)>`), giving a clear yet flexible structure without enforcing a specific serialization format.
 - **Off-chain metadata** (`token_uri`):
   - Additionally, CEP-95 supports traditional off-chain metadata retrieval via a URI. Contracts can provide a URI linking to external JSON metadata (e.g., hosted on IPFS or HTTP servers). This off-chain approach reduces on-chain storage costs and allows easier updates of non-critical metadata fields.
+
+When metadata is created or updated, the contract emits a `MetadataUpdate` event, ensuring that indexers and clients are immediately informed and can cache or display this data without additional lookups.
 
 Consumers, such as wallets, explorers, and marketplaces, benefit from this hybrid approach by first attempting to retrieve on-chain metadata for immediate trustworthiness and availability. If on-chain metadata is unavailable or limited, clients can fall back on the provided metadata URI. This method offers the best balance between efficiency, flexibility, trust, and ease of use, ensuring a smooth and user-friendly NFT experience within the Casper ecosystem.
 
@@ -211,18 +213,10 @@ pub trait CEP95TokenMetadata {
     /// @param token_id - The NFT ID.
     /// @return Option<Vec<(String, String)>> - optional metadata stored as a list of key-value pairs.
     fn token_metadata(&self, token_id: U256) -> Option<Vec<(String, String)>>;
-    
-    /// Returns a URI pointing to metadata for a specific token.
-    ///
-    /// @param token_id - The NFT ID.
-    /// @return Option<String> - A URI pointing to metadata if available.
-    fn token_uri(&self, token_id: U256) -> Option<String>;
 }
 ```
 
-### CEP-95 On-chain Metadata Schema
-
-When providing NFT metadata (either through the off-chain URI from token_uri or directly stored on-chain via token_metadata), it's highly recommended to follow this conventional JSON schema:
+### On-chain Token Metadata Schema
 
 ```json
 {
@@ -240,10 +234,26 @@ When providing NFT metadata (either through the off-chain URI from token_uri or 
         "asset_uri": {
             "type": "string",
             "description": "A URI pointing to a resource with mime type image/* representing the asset to which this NFT represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive."
-        },
+        }
+    },
+    "required": []
+}
+```
+
+### Offchain Token Metadata Schema
+
+```json
+{
+    "title": "Casper NFT Metadata",
+    "type": "object",
+    "properties": {
         "token_uri": {
           "type": "string",
           "description": "A URI pointing to additional metadata stored off-chain (typically JSON format compatible with this schema)."
+        },
+        "checksum": {
+          "type": "string",
+          "description": "Checksum hash of the offchain metadata stored via token_uri"
         }
     },
     "required": []
@@ -348,6 +358,7 @@ Once an event (such as minting) is detected:
 [conclusion]: #conclusion
 
 The Casper NFT Standard (inspired by ERC-721) aims to bring uniformity and interoperability to NFT development on Casper. It provides simple methods for handling tokens, approvals, and optional metadata, making it easier for wallets, marketplaces, and indexers to work with NFTs. This proposal helps build a connected and thriving NFT ecosystem on Casper.
+
 ## Appendix
 
 [appendix]: #appendix
@@ -355,10 +366,10 @@ The Casper NFT Standard (inspired by ERC-721) aims to bring uniformity and inter
 ### Changelog
 
 - v1.0: Initial draft including CEP-95 and CEP95Metadata interfaces, event structures, and backend indexing guidelines.
-- v1.1: Updated approach with optional Token Metadata. Added examples for tracking offchain and onchain metadata. Added recommendations for storing contract metadata. Added ref with recommendation to utilise CEP-96 for contract metadata.
 
 ### References
 
 - [ERC-721 Standard on Ethereum](https://eips.ethereum.org/EIPS/eip-721)
+- [CEP-96 Contract Metadata Standard](https://github.com/casper-network/ceps/pull/96)
 - [Casper Documentation (for contract APIs and types)](https://docs.casper.network/concepts/smart-contracts)
 - [Example proposals and CEPs from the Casper community](https://github.com/make-software/casper-ceps/tree/master/text)
--- a/text/0095-nft-standard.md
+++ b/text/0095-nft-standard.md
@@ -159,14 +159,6 @@ pub trait CEP95Receiver {
 }
 ```
 
-### Contract Metadata Standard CEP-96
-
-To improve the integrity, discoverability, and user experience within the Casper ecosystem, implementers of the CEP-95 NFT Standard are strongly encouraged to also adopt the **CEP-96 Casper On-chain Contract Metadata Standard**.
-
-By implementing these metadata standard, CEP-95 contracts become easily identifiable, understandable, and trustworthy to end-users, explorers, wallets, and marketplaces. The adoption of the On-chain Contract Metadata Standard enhances the overall integrity and usability of the Casper blockchain ecosystem and helps build user trust.
-
-We strongly recommend incorporating this metadata standard alongside CEP-95 implementations to contribute positively to ecosystem consistency, transparency, and user-friendliness.
-
 ### Contract symbol metadata
 
 To keep contract metadata storage consistent it is strongly recommended to follow the CEP-96 standard.
--- a/text/0095-nft-standard.md
+++ b/text/0095-nft-standard.md
@@ -69,7 +69,7 @@ pub enum CEP95Event {
 /// Casper-compatible NFT interface
 pub trait CEP95 {
     /// Returns an optional short symbol or abbreviation for the NFT collection.
-    fn symbol(&self) -> Option<String>;
+    fn symbol(&self) -> String;
 
     /// Returns the number of NFTs owned by a given account or contract
     ///
@@ -89,7 +89,7 @@ pub trait CEP95 {
     /// @param to - The new owner.
     /// @param token_id - The NFT ID.
     /// @param data - Optional payload to pass to a receiving contract.
-    fn safe_transfer_from(&mut self, from: Key, to: Key, token_id: U256, data: Option<Vec<u8>>);
+    fn safe_transfer_from(&mut self, from: Key, to: Key, token_id: U256, data: Option<Bytes>);
 
     /// Transfers the ownership of an NFT without checking the recipient contract.
     ///
@@ -150,26 +150,24 @@ To keep contract metadata storage consistent it is strongly recommended to follo
 
 1. Install/Constructor Phase:
     ```rust
-    let symbol_arg: Option<String> = runtime::get_named_arg("symbol");
-    if let Some(symbol_value) = symbol_arg {
-        // Create a new URef for the symbol and place it as a named key
-        let symbol_uref = storage::new_uref(symbol_value);
-        runtime::put_key("symbol", symbol_uref.into());
-    }
+    let symbol: String = runtime::get_named_arg("symbol");
+    
+    let symbol_uref = storage::new_uref(symbol);
+    runtime::put_key("symbol", symbol_uref.into());
     ```
 2. Interface Method:
     ```rust
     #[no_mangle]
     pub extern "C" fn symbol() {
-        let symbol: Option<String> = match runtime::get_key("symbol") {
-            Some(key) => {
-                let symbol_uref = key.into_uref().expect("symbol key is not a URef");
-                storage::read(symbol_uref).expect("unable to read symbol from storage")
-            }
-            None => None,
-        };
+        let key = runtime::get_key("symbol").expect("symbol key not found");
+
+        let symbol_uref = key.into_uref().expect("symbol key is not a URef");
+        let symbol: String = storage::read(symbol_uref)
+            .unwrap_or_revert()       // read() -> Result<Option<T>, _>
+            .unwrap_or_revert();      // Option<T> -> T (revert if None)
+    
         runtime::ret(CLValue::from_t(symbol).unwrap_or_revert());
-    } 
+    }
     ```
 
 ## Token Metadata Extension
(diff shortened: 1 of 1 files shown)


text/0093-nft-standard.md
```diff
@@ -0,0 +1,260 @@
...
+#![no_std]
+
+use casper_contract::contract_api::{runtime, storage};
+use casper_types::{Key, URef, U256, AccountHash, ContractHash};
+
+/// Enum defining events that should be emitted by NFT operations
+#[derive(Debug)]
+pub enum CEP95Event {
+    /// Emitted when an NFT is transferred, minted (from=None), or burned (to=None)
+    Transfer { from: Option<AccountHash>, to: Option<AccountHash>, token_id: U256 },
```
ihor, line 53: 1. `from`  and `to` should be of type `Key` since those could be also contract package hashes
2. Let's have separate `Mint` and `Burn` events. Those are more transparent and already were present in the existing CEP-47 and CEP-78
3. @davidatwhiletrue @zie1ony what's your opinion about integer vs hash ids?

text/0093-nft-standard.md
```diff
@@ -0,0 +1,260 @@
...
+
+    /// Emitted when a specific NFT is approved to an account
+    Approval { owner: AccountHash, approved: Option<AccountHash>, token_id: U256 },
```
ihor, line 56: 1. `ApprovalRevoked` event is missing
2. I propose to rename `approved` to `spender` since it was the naming used in CEP-47 and CEP-78
2. Both`owner` and `spender` should be of type `Key`

text/0093-nft-standard.md
```diff
@@ -0,0 +1,260 @@
...
+    ApprovalForAll { owner: AccountHash, operator: AccountHash, approved: bool },
+}
+
+/// Casper-compatible NFT interface
+pub trait CEP95 {
+    /// Returns the number of NFTs owned by a given account
+    ///
+    /// @param owner - The account to query.
+    /// @return U256 - The number of NFTs owned.
+    fn balance_of(&self, owner: AccountHash) -> U256;
```
ihor, line 68: Basically, in all methods `AccountHash` should be replaced with `Key` since tokens can be owned also by a contract.

text/0093-nft-standard.md
```diff
@@ -0,0 +1,260 @@
...
+    /// @return Option<AccountHash> - The owner if it exists, else None.
+    fn owner_of(&self, token_id: U256) -> Option<AccountHash>;
+
+    /// Transfers the ownership of an NFT and performs a recipient check.
+    ///
+    /// @param from - The current owner of the NFT.
+    /// @param to - The new owner.
+    /// @param token_id - The NFT ID.
+    /// @param data - Optional payload to pass to a receiving contract.
+    fn safe_transfer_from(&mut self, from: AccountHash, to: AccountHash, token_id: U256, data: Option<Vec<u8>>);
```
ihor, line 82: @davidwhiletrue @zie1ony what's your opinion on this one?

text/0093-nft-standard.md
```diff
@@ -0,0 +1,260 @@
...
+    /// @param from - The current owner of the NFT.
+    /// @param to - The new owner.
+    /// @param token_id - The NFT ID.
+    fn transfer_from(&mut self, from: AccountHash, to: AccountHash, token_id: U256);
+
+    /// Approves another account to transfer a specific NFT.
+    ///
+    /// @param to - The account that will be granted approval.
+    /// @param token_id - The NFT ID.
+    fn approve(&mut self, to: Option<AccountHash>, token_id: U256);
```
ihor, line 95: I propose to change `to` to `spender` since it was the term used in CEP-47 and CEP-78, and also to match the event parameter

text/0093-nft-standard.md
```diff
@@ -0,0 +1,260 @@
...
+    /// Revokes operator approval for all of the caller's NFTs.
+    ///
+    /// @param operator - The operator address whose approval is to be revoked.
+    fn revoke_approval_for_all(&mut self, operator: AccountHash);
+
+    /// Gets the approved account for a specific NFT.
+    ///
+    /// @param token_id - The NFT ID.
+    /// @return Option<AccountHash> - Approved account if one exists.
+    fn get_approved(&self, token_id: U256) -> Option<AccountHash>;
```
ihor, line 116: Let's put this method right below the `revoke_approval`

text/0093-nft-standard.md
```diff
@@ -0,0 +1,260 @@
...
+
+    /// Checks if an operator is approved to manage all NFTs of the owner.
+    ///
+    /// @param owner - The NFT owner's address.
+    /// @param operator - The operator to check.
+    /// @return bool - True if the operator is approved for all NFTs, false otherwise.
+    fn is_approved_for_all(&self, owner: AccountHash, operator: AccountHash) -> bool;
```
ihor, line 123: I'd consider naming it `has_approved_for_all` like "has the owner approved operator"

text/0093-nft-standard.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Metadata Interface: CEP95Metadata
+
+```rust
+pub trait CEP95Metadata {
+    /// Returns a descriptive name for a collection of NFTs.
+    fn name(&self) -> String;
+
+    /// Returns a short symbol or abbreviation for the NFT collection.
+    fn symbol(&self) -> String;
```
ihor, line 144: @davidatwhiletrue @zie1ony do you know any utility for a symbol for NFTs?

text/0093-nft-standard.md
ihor: Per @zie1ony comment something like this should be described in this proposal as well:
```
/// @dev Throws unless `msg.sender` is the current owner, an authorized
    ///  operator, or the approved address for this NFT. Throws if `_from` is
    ///  not the current owner. Throws if `_to` is the zero address. Throws if
    ///  `_tokenId` is not a valid NFT. When transfer is complete, this function
    ///  checks if `_to` is a smart contract (code size > 0). If so, it calls
    ///  `onERC721Received` on `_to` and throws if the return value is not
    ///  `bytes4(keccak256("onERC721Received(address,address,uint256,bytes)"))`.
```

text/0093-nft-standard.md
```diff
@@ -0,0 +1,260 @@
...
+
+    /// Returns a URI pointing to metadata for a specific token.
+    ///
+    /// @param token_id - The NFT ID.
+    /// @return Option<String> - A URI pointing to JSON metadata if available.
+    fn token_uri(&self, token_id: U256) -> Option<String>;
```
ihor, line 150: Maybe let's have a `get_token_metadata` method that return the on-chain metadata and have a separate standard for the metadata structure with the optional following fields:
- `name`
- `description`
- `asset_uri`
- `token_uri`

text/0093-nft-standard.md
davidatwhiletrue: @ihor I would allow both, integer and string token identifiers.

text/0095-nft-standard.md
```diff
@@ -0,0 +1,340 @@
...
+    /// @return U256 - The number of NFTs owned.
+    fn balance_of(&self, owner: Key) -> U256;
+
+    /// Returns the owner of a specific NFT.
+    ///
+    /// @param token_id - The ID of the NFT.
+    /// @return Option<Key> - The owner if it exists, else None.
+    fn owner_of(&self, token_id: U256) -> Option<Key>;
+
+    /// Transfers the ownership of an NFT and performs a recipient check.
```
davidatwhiletrue, line 88: I'd rearrange the steps. First the check, then the transfer.

text/0095-nft-standard.md
```diff
@@ -0,0 +1,340 @@
...
+    fn safe_transfer_from(&mut self, from: Key, to: Key, token_id: U256, data: Option<Vec<u8>>);
+
+    /// Transfers the ownership of an NFT without checking the recipient contract.
+    ///
+    /// @param from - The current owner of the NFT.
+    /// @param to - The new owner.
+    /// @param token_id - The NFT ID.
+    fn transfer_from(&mut self, from: Key, to: Key, token_id: U256);
+
+    /// Approves another account to transfer a specific NFT.
```
davidatwhiletrue, line 103: account or contract

text/0095-nft-standard.md
```diff
@@ -0,0 +1,340 @@
...
+    /// @param to - The account that will be granted approval.
+    /// @param token_id - The NFT ID.
+    fn approve(&mut self, spender: Key, token_id: U256);
+
+    /// Revokes approval for a specific NFT.
+    ///
+    /// @param token_id - The NFT ID.
+    fn revoke_approval(&mut self, token_id: U256);
+
+    /// Gets the approved account for a specific NFT.
```
davidatwhiletrue, line 114: account or contract

text/0095-nft-standard.md
```diff
@@ -0,0 +1,340 @@
...
+
+    /// Returns a short symbol or abbreviation for the NFT collection.
+    fn symbol(&self) -> String;
+}
+```
+
+### Token Offchain Metadata Interface: CEP95TokenOffchainMetadata
+
+```rust
+/// Optional interface for retrieving on-chain token metadata
```
davidatwhiletrue, line 160: off-chain

text/0095-nft-standard.md
```diff
@@ -0,0 +1,340 @@
...
+pub trait CEP95TokenOffchainMetadata {
+    /// Returns a URI pointing to metadata for a specific token.
+    ///
+    /// @param token_id - The NFT ID.
+    /// @return Option<String> - A URI pointing to JSON metadata if available.
```
davidatwhiletrue, line 165: maybe we don't need to mention JSON here.

text/0095-nft-standard.md
```diff
@@ -0,0 +1,340 @@
...
+
+### Token Onchain Metadata Interface: CEP95TokenOffchainMetadata
+
+```rust
+#[derive(Debug)]
+pub enum CEP95MetadataEvent {
+    /// Emitted whenever on-chain metadata for a token is created or updated.
+    ///
+    /// @param token_id - The ID of the NFT whose metadata was updated
+    /// @param metadata - The raw metadata (e.g., JSON bytes)
```
davidatwhiletrue, line 180: remove `(e.g., JSON bytes)`?

text/0095-nft-standard.md
```diff
@@ -0,0 +1,340 @@
...
+        metadata: Vec<(String, String)>,
+    },
+}
+
+/// Optional interface for retrieving on-chain token metadata
+pub trait CEP95TokenOnchainMetadata {
+    /// Returns raw metadata for a given token ID.
+    ///
+    /// @param token_id - The NFT ID.
+    /// @return Option<Vec<(String, String)>> - Encoded metadata blob (e.g., JSON-encoded bytes, CBOR, etc.)
```
davidatwhiletrue, line 192: maybe `a list of key-value pairs` instead of `encoded...` ?

text/0095-nft-standard.md
```diff
@@ -0,0 +1,371 @@
+# Casper NFT Standard Proposal
+
+Casper NFT Standard Proposal (Inspired by ERC-721 & adapted for the Casper ecosystem)
+
+## Summary
+
+CEP PR: [casperlabs/ceps#0095](https://github.com/casper-network/ceps/pull/95)
```
ihor, line 7: Let's remove the link to the PR here. It also mentions Casperlabs here.

text/0095-nft-standard.md
```diff
@@ -0,0 +1,371 @@
...
+### Recommended Adoption of the Casper On-chain Contract Metadata Standard CEP96
+
+To improve the integrity, discoverability, and user experience within the Casper ecosystem, implementers of the CEP95 NFT Standard are strongly encouraged to also adopt the **CEP96 Casper On-chain Contract Metadata Standard**.
+
+This standard provides a clear, unified way for contracts to publish immutable, verifiable metadata about themselves directly on-chain, including:
+
+- **Contract Name** (`name`)
+- **Description** (`description`)
+- **Icon URI** (`icon_uri`)
+- **Project URI** (`project_uri`)
```
ihor, line 148: Let's not duplicate the details here

text/0095-nft-standard.md
```diff
@@ -0,0 +1,371 @@
...
+    },
+}
+
+/// Optional interface for retrieving token metadata
+pub trait CEP95TokenMetadata {
+    /// Returns metadata for a given token ID.
+    ///
+    /// @param token_id - The NFT ID.
+    /// @return Option<Vec<(String, String)>> - optional metadata stored as a list of key-value pairs.
+    fn token_metadata(&self, token_id: U256) -> Option<Vec<(String, String)>>;
```
ihor, line 219: I'd call this method `get_token_metadata` to keep it consistent with `get_approved` and also because Casper conrtracts aren't limited to Rust only.

text/0095-nft-standard.md
```diff
@@ -0,0 +1,371 @@
...
+    
+    /// Returns a URI pointing to metadata for a specific token.
+    ///
+    /// @param token_id - The NFT ID.
+    /// @return Option<String> - A URI pointing to metadata if available.
+    fn token_uri(&self, token_id: U256) -> Option<String>;
```
ihor, line 225: I'd remove this method from here because it implies the token should have an external URI which is not always the case. Additionally, it doesn't communicate that it's for offchain metadata well enough. 

Instead let's add a section called `Offchain metadata" that would describe the approach and provide an on-chain metadata schema:
```
{
    "token_uri": ...,
    "checksum": ...,
}
```

text/0095-nft-standard.md
```diff
@@ -0,0 +1,371 @@
...
+}
+```
+
+### Conventional Metadata Schema
```
ihor, line 229: Let's call it "CEP95 On-chain Metadata Schema"

text/0095-nft-standard.md
```diff
@@ -0,0 +1,371 @@
...
+- **Icon URI** (`icon_uri`)
+- **Project URI** (`project_uri`)
+
+By implementing these metadata fields, CEP95 contracts become easily identifiable, understandable, and trustworthy to end-users, explorers, wallets, and marketplaces. The adoption of the On-chain Contract Metadata Standard enhances the overall integrity and usability of the Casper blockchain ecosystem and helps build user trust.
+
+We strongly recommend incorporating this metadata standard alongside CEP95 implementations to contribute positively to ecosystem consistency, transparency, and user-friendliness.
+
+### Recommended Storage for contract data (symbol, etc.)
+
+According to the CEP96 standard, it is recommended to store contract metadata as a form of contract level named keys using `Urefs`.
```
ihor, line 156: `According to the CEP96 standard` is not needed here since we are looking at the standard iteself

text/0095-nft-standard.md
```diff
@@ -0,0 +1,371 @@
...
+To improve the integrity, discoverability, and user experience within the Casper ecosystem, implementers of the CEP95 NFT Standard are strongly encouraged to also adopt the **CEP96 Casper On-chain Contract Metadata Standard**.
+
+This standard provides a clear, unified way for contracts to publish immutable, verifiable metadata about themselves directly on-chain, including:
+
+- **Contract Name** (`name`)
+- **Description** (`description`)
+- **Icon URI** (`icon_uri`)
+- **Project URI** (`project_uri`)
+
+By implementing these metadata fields, CEP95 contracts become easily identifiable, understandable, and trustworthy to end-users, explorers, wallets, and marketplaces. The adoption of the On-chain Contract Metadata Standard enhances the overall integrity and usability of the Casper blockchain ecosystem and helps build user trust.
```
ihor, line 150: Let's use `-` in standard names since it's an already established convention: `CEP-95`

text/0095-nft-standard.md
```diff
@@ -0,0 +1,371 @@
...
+
+This standard defines a unified API for implementing non-fungible tokens (NFTs) on the Casper network. It introduces core functionality for tracking and transferring NFTs in a consistent and interoperable way.
+
+We have taken into account use cases where NFTs are owned directly by individuals, as well as scenarios where ownership and transfers are managed by third-party entities such as marketplaces, wallets, or custodial services ("operators"). NFTs on Casper can represent ownership of both digital and physical assets, and the range of potential applications continues to grow:
+
+- Physical assets — real estate, unique artwork 
+- Digital collectibles — rare digital items, game assets, virtual companions 
+- Obligations and non-traditional assets — debt instruments, legal contracts, responsibilities
+
+Each NFT is uniquely identifiable, and as such, must be tracked individually. Unlike fungible tokens, NFTs are not interchangeable — every item is distinct, and this standard ensures that uniqueness and ownership are verifiable and transferable on-chain.
```
ihor, line 17: The section above is copypasted from ERC-721. Let's provide the real summary and background behind this CEP. Something like:

```
This standard provides a non-fungible token (NFT) contract API that is consistent with the existing [CEP-18](link here) fungible token standard. It aims to replace [CEP-47](link here) and [CEP-78](link here), which have flaws that complicate their support in the ecosystem. CEP-47 lacks operator approval functionality, which makes it less convenient to integrate with NFT marketplaces. CEP-78 rather describes a configurable NFT contract that can modify its behavior to various needs, rather than a plain NFT contract API.

This standard is aligned with Ethereum's [ERC-721](link here), but makes adjustments relevant for the Casper Ecosystem. Similarly to ERC-721, this standard can be used to represent a various range of tokenized assets:
 - Physical assets — real estate, unique artwork 
 - Digital collectibles — rare digital items, game assets, virtual companions 
 - Obligations and non-traditional assets — debt instruments, legal contracts, responsibilities
 
Each NFT is uniquely identifiable, and as such, must be tracked individually. Unlike fungible tokens, NFTs are not interchangeable — every item is distinct, and this standard ensures that uniqueness and ownership are verifiable and transferable on-chain.
```

text/0095-nft-standard.md
```diff
@@ -0,0 +1,371 @@
...
+A standard interface for non-fungible tokens (NFTs) on the Casper network enables wallets, marketplaces, decentralized applications, and other ecosystem tools to interact seamlessly with any compliant NFT contract. The lessons learned from Ethereum’s ERC-721 standard show that a consistent interface increases interoperability and speeds up adoption. On Casper, this standard will ensure that every NFT, whether tracking digital collectibles or physical assets, is uniquely identifiable and transferable.
+
+## Objective
+
+[objective]: #objective
+
+The objective of this proposal is to define a Casper-compatible NFT standard that supports:
+
+- Tracking ownership and metadata of NFTs. 
+- Transferring NFTs safely between accounts.
```
ihor, line 31: "between accounts and contracts"

text/0095-nft-standard.md
ihor: @zie1ony @davidatwhiletrue I propose to move this method in the main interface, suggesting that NFTs should always have metadata. What do you think?

text/0095-nft-standard.md
```diff
@@ -0,0 +1,371 @@
...
+[motivation]: #motivation
+A standard interface for non-fungible tokens (NFTs) on the Casper network enables wallets, marketplaces, decentralized applications, and other ecosystem tools to interact seamlessly with any compliant NFT contract. The lessons learned from Ethereum’s ERC-721 standard show that a consistent interface increases interoperability and speeds up adoption. On Casper, this standard will ensure that every NFT, whether tracking digital collectibles or physical assets, is uniquely identifiable and transferable.
+
+## Objective
+
+[objective]: #objective
+
+The objective of this proposal is to define a Casper-compatible NFT standard that supports:
+
+- Tracking ownership and metadata of NFTs.
```
ihor, line 30: The metadata part kind of suggests that the metadata interface should be a part of the main interface

text/0095-nft-standard.md
```diff
@@ -0,0 +1,371 @@
...
+## Objective
+
+[objective]: #objective
+
+The objective of this proposal is to define a Casper-compatible NFT standard that supports:
+
+- Tracking ownership and metadata of NFTs. 
+- Transferring NFTs safely between accounts. 
+- Managing per-token approvals as well as operator-level approvals.
+- (Optionally) exposing metadata (name, symbol, token URI) so that applications can easily display NFT details.
```
ihor, line 33: I'd remove or change this line. NFT metadata is about describing the underlying asset, rather than having a possibility to display it. 

The display problem is handled by providing a recommended convention for standard NFT metadata properties which can be found below in the standard.

text/0095-nft-standard.md
```diff
@@ -0,0 +1,364 @@
...
+
+### Recommended Adoption of the Casper On-chain Contract Metadata Standard CEP96
+
+To improve the integrity, discoverability, and user experience within the Casper ecosystem, implementers of the CEP-95 NFT Standard are strongly encouraged to also adopt the **CEP96 Casper On-chain Contract Metadata Standard**.
+
+By implementing these metadata standard, CEP-95 contracts become easily identifiable, understandable, and trustworthy to end-users, explorers, wallets, and marketplaces. The adoption of the On-chain Contract Metadata Standard enhances the overall integrity and usability of the Casper blockchain ecosystem and helps build user trust.
+
+We strongly recommend incorporating this metadata standard alongside CEP-95 implementations to contribute positively to ecosystem consistency, transparency, and user-friendliness.
+
+### Recommended Storage for contract data (symbol, etc.)
```
ihor, line 145: Let's shorten to "Contract symbol metadata"

text/0095-nft-standard.md
```diff
@@ -0,0 +1,364 @@
...
+    
+    /// Returns a URI pointing to metadata for a specific token.
+    ///
+    /// @param token_id - The NFT ID.
+    /// @return Option<String> - A URI pointing to metadata if available.
+    fn token_uri(&self, token_id: U256) -> Option<String>;
+}
+```
+
+### CEP-95 On-chain Metadata Schema
```
ihor, line 223: I'd name it "On-chain token metadata" and would add a separate "Off-chain token metadata"

text/0095-nft-standard.md
```diff
@@ -0,0 +1,364 @@
...
+- Transferring NFTs safely between accounts and contracts.
+- Managing per-token approvals as well as operator-level approvals.
+- (Optionally) exposing metadata (name, symbol, token URI) that describes the underlying asset details.
+
+This proposal takes into account a possibility of off-chain indexing of NFTs by specifying interfaces of the events that can be used to track the contract activity.
+## Specification
+
+[specification]: #specification
+
+### Core Casper NFT Interface
```
ihor, line 39: Let's shorten "NFT Contract Interface"

text/0095-nft-standard.md
```diff
@@ -0,0 +1,364 @@
...
+    /// Revokes approval for a specific NFT.
+    ///
+    /// @param token_id - The NFT ID.
+    fn revoke_approval(&mut self, token_id: U256);
+
+    /// Gets the approved account or contract for a specific NFT.
+    ///
+    /// @param token_id - The NFT ID.
+    /// @return Option<Key> - Approved spender account if one exists.
+    fn get_approved(&self, token_id: U256) -> Option<Key>;
```
ihor, line 116: `approved_for` if we decide to keep getters w/o the `get` prefix

text/0095-nft-standard.md
```diff
@@ -0,0 +1,364 @@
+# Casper NFT Standard Proposal
+
+Casper NFT Standard Proposal (Inspired by ERC-721 & adapted for the Casper ecosystem)
+
+## Summary
+
+CEP PR: [casperlabs/ceps#0095]
```
ihor, line 7: Change `casperlabs` to `casper-network`

text/0095-nft-standard.md
```diff
@@ -0,0 +1,364 @@
...
+
+The Casper NFT Standard (inspired by ERC-721) aims to bring uniformity and interoperability to NFT development on Casper. It provides simple methods for handling tokens, approvals, and optional metadata, making it easier for wallets, marketplaces, and indexers to work with NFTs. This proposal helps build a connected and thriving NFT ecosystem on Casper.
+## Appendix
+
+[appendix]: #appendix
+
+### Changelog
+
+- v1.0: Initial draft including CEP-95 and CEP95Metadata interfaces, event structures, and backend indexing guidelines.
+- v1.1: Updated approach with optional Token Metadata. Added examples for tracking offchain and onchain metadata. Added recommendations for storing contract metadata. Added ref with recommendation to utilise CEP-96 for contract metadata.
```
ihor, line 358: Let's keep only v1.0 until it's merged (meaning the first version is accepted)

text/0095-nft-standard.md
```diff
@@ -0,0 +1,364 @@
...
+
+### References
+
+- [ERC-721 Standard on Ethereum](https://eips.ethereum.org/EIPS/eip-721)
```
ihor, line 362: Let's also reference to CEP-96 here

text/0095-nft-standard.md
```diff
@@ -0,0 +1,375 @@
...
+        "asset_uri": {
+            "type": "string",
+            "description": "A URI pointing to a resource with mime type image/* representing the asset to which this NFT represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive."
+        }
+    },
+    "required": []
+}
+```
+
+### Offchain Token Metadata Schema
```
ihor, line 243: Let's call it "Off-chain" since we named the prev section "On-chain"

text/0095-nft-standard.md
```diff
@@ -0,0 +1,375 @@
...
+
+The Casper NFT Standard (inspired by ERC-721) aims to bring uniformity and interoperability to NFT development on Casper. It provides simple methods for handling tokens, approvals, and optional metadata, making it easier for wallets, marketplaces, and indexers to work with NFTs. This proposal helps build a connected and thriving NFT ecosystem on Casper.
+
+## Appendix
+
+[appendix]: #appendix
+
+### Changelog
+
+- v1.0: Initial draft including CEP-95 and CEP95Metadata interfaces, event structures, and backend indexing guidelines.
```
ihor, line 368: It's `CEP95` here. Let's use `code` formating for interfaces and plain text for standard names

zie1ony (changes requested):
For all metadata structs add:
- name of the named key data lives.
- CLType of the value.

For all dictionary based data structures add:
- name of the named key for the dictionary,
- how to create string with the key,
- CLType of the value.

For events it should be clearly defined that:
- We follow CES events.
- In future we might switch to native events, and then this spec will be updated.

text/0095-nft-standard.md
```diff
@@ -0,0 +1,375 @@
...
+pub trait CEP95TokenMetadata {
+    /// Returns metadata for a given token ID.
+    ///
+    /// @param token_id - The NFT ID.
+    /// @return Option<Vec<(String, String)>> - optional metadata stored as a list of key-value pairs.
+    fn token_metadata(&self, token_id: U256) -> Option<Vec<(String, String)>>;
+}
+```
+
+### On-chain Token Metadata Schema
```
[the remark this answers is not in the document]
ihor: @zie1ony I think that's what specified in the interface above.  I think @Volodymyr-Kuchinskyi just used JSON schema to describe the reecommended most commonly used names.

text/0095-nft-standard.md
```diff
@@ -0,0 +1,403 @@
...
+    },
+}
+
+/// Optional interface for retrieving token metadata
+pub trait CEP95TokenMetadata {
+    /// Returns metadata for a given token ID.
+    ///
+    /// @param token_id - The NFT ID.
+    /// @return Vec<(String, String)> - metadata stored as a list of key-value pairs.
+    fn token_metadata(&self, token_id: U256) -> Vec<(String, String)>;
```
ihor, line 241: @zie1ony @davidatwhiletrue I propose to move the method above in the main standard interface. If we worry that thaere will be use-cases for tokens w/o metadata, we could wrap the return type in `Option`. My motivation is the following:
1. It enforces the metadata type (a map), which is better for indexing and displaying the assets in network explorers and wallets.
2. It collocates the metadata related event together with the rest of the events, thus makes it easier to see it. I suspect that many readers will focus mostly on the main standard interface.

@Volodymyr-Kuchinskyi please don't force push, because it's killing my comments

text/0095-nft-standard.md
```diff
@@ -0,0 +1,375 @@
...
+### CEP95MetadataEvent Enum
+
+```rust
+pub enum CEP95MetadataEvent {
+    /// Emitted whenever on-chain metadata for a token is created or updated.
+    ///
+    /// @param token_id - The ID of the NFT whose metadata was updated
+    /// @param metadata - The metadata represented in key/value pairs
+    MetadataUpdate {
+        token_id: U256,
```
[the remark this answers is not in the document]
davidatwhiletrue: @ihor, while it makes things easier, I don't like to store duplicate data. If I'm not mistaken, we can add the block height to the event payload. That could help to locate changes quickly for someone reading historical data through events.

text/0095-nft-standard.md
davidatwhiletrue: I think it's a good idea. I'd also use `Option<>` in the return type.

zie1ony: `Option` doesn't really make sense here. It's better to return empty vector, even for token id that doesn't exist.

text/0095-nft-standard.md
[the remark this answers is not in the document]
zie1ony: We store duplicated data for all transfers and approvals. I also don't think "locate changes quickly for someone reading historical data through events" is an requirement we need.

text/0095-nft-standard.md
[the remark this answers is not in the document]
zie1ony: ok

text/0093-nft-standard.md
zie1ony: Having integers and hashes as ids was super annoying while working with cep78. `U256` is large enough to have hash inside it if you want.

text/0095-nft-standard.md
```diff
@@ -0,0 +1,375 @@
...
+    ApprovalForAll { owner: Key, operator: Key },
+    
+    /// Emitted when an operator approval is revoked for all NFTs of an owner
+    RevokeApprovalForAll { owner: Key, operator: Key },
+}
+
+/// Casper-compatible NFT interface
+pub trait CEP95 {
+    /// Returns an optional short symbol or abbreviation for the NFT collection.
+    fn symbol(&self) -> Option<String>;
```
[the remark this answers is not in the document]
zie1ony: I reviewed CEP96 as well. "Contract Metadata" and "Token Metadata" are two different topic, that we should not mix. "Token Metadata" are responsible for delivering you metadata about token, like token name. It can be different to what the contract name is.
For example I can have contract that is both a token and a multisig-wallet. Then token name is "My multisig token name" and contract name would be "wallet". We are also dealing with referencing ERC721 and this makes all the confusions:
- It would be better if the naming was `token_symbol` and `token_name`. Than would make it more explicit and I heard this many times from angry Solidity devs that would like to have both ERC20 and ERC721 in a single contract. But ERC721 is already adopted and I prefer to follow this standard as is, so we create as little confusion as possible for new smart contract developers coming to Casper.
- I would also leave it as a standard elements in CEP95, for the compatibility reasons with tokens ported via bridges (that we will have at some point in future). It's better if we overlap with ERC721 as possible, so there is no ambiguity when porting data from one blockchain to another.

zie1ony (approved):
All good now :)

text/0095-nft-standard.md
ihor: @Volodymyr-Kuchinskyi it seems that the consensus  here is to merge the NFT metadata interface into the main contract interface. Let's do that

text/0095-nft-standard.md
```diff
@@ -0,0 +1,387 @@
...
+    /// Transfers the ownership of an NFT without checking the recipient contract.
+    ///
+    /// @param from - The current owner of the NFT.
+    /// @param to - The new owner.
+    /// @param token_id - The NFT ID.
+    fn transfer_from(&mut self, from: Key, to: Key, token_id: U256);
+
+    /// Approves another account or contract to transfer a specific NFT.
+    ///
+    /// @param to - The account that will be granted approval.
```
davidatwhiletrue: `@param spender` instead of `@param to`.
The account or contract...

text/0095-nft-standard.md
```diff
@@ -0,0 +1,387 @@
...
+
+    /// Revokes approval for a specific NFT.
+    ///
+    /// @param token_id - The NFT ID.
+    fn revoke_approval(&mut self, token_id: U256);
+
+    /// Gets the approved account or contract for a specific NFT.
+    ///
+    /// @param token_id - The NFT ID.
+    /// @return Option<Key> - Approved spender account if one exists.
```
davidatwhiletrue: Approved spender account or contract if one exists.

text/0095-nft-standard.md
```diff
@@ -0,0 +1,387 @@
...
+
+### Hybrid Approach for Token Metadata in CEP-95
+
+[hybrid-approach-for-token-metadata-in-cep-95]: #hybrid-approach-for-token-metadata-in-cep-95
+
+The CEP-95 standard provides a flexible, hybrid approach for NFT token metadata, allowing developers and consumers to leverage **both on-chain and off-chain metadata sources**. Metadata is stored as key-value pairs (`Vec<(String, String)>`), giving a clear yet flexible structure without enforcing a specific serialization format.
+
+- **On-chain metadata**:
+  - Contract stores all its NFT's metadata onchain.
+  - Typically, it its dictionary by `token_metadata` Uref on a contract level where key is a `token_id` and value is a metadata itself.
```
davidatwhiletrue: `Typically, it is stored in a dictionary under ...` ?

text/0095-nft-standard.md
```diff
@@ -0,0 +1,387 @@
...
+### Changelog
+
+- v1.0: Initial draft including CEP-95 and `CEP95Metadata` interfaces, event structures, and backend indexing guidelines.
+
+### References
+
+- [ERC-721 Standard on Ethereum](https://eips.ethereum.org/EIPS/eip-721)
+- [CEP-96 Contract Metadata Standard](https://github.com/casper-network/ceps/pull/96)
+- [Casper Documentation (for contract APIs and types)](https://docs.casper.network/concepts/smart-contracts)
+- [Example proposals and CEPs from the Casper community](https://github.com/make-software/casper-ceps/tree/master/text)
```
davidatwhiletrue: should this be `https://github.com/casper-network/ceps/tree/master/text`?

text/0095-nft-standard.md
```diff
@@ -0,0 +1,387 @@
...
+    fn token_metadata(&self, token_id: U256) -> Vec<(String, String)>;
+}
+
+/// Receiver interface
+pub trait CEP95Receiver {
+    /// Called after a `safe_transfer_from` completes its internal state update.
+    /// MUST return `true` to signal acceptance; returning `false` or reverting
+    /// causes the entire transfer to roll back.
+    ///
+    /// @param operator The account (EOA or contract) that invoked
```
davidatwhiletrue: EOA is not defined. Maybe `The account or contract that ...`?

text/0095-nft-standard.md
```diff
@@ -0,0 +1,387 @@
...
+pub trait CEP95 {
+    /// Returns a name of the NFT token/collection.
+    fn name(&self) -> String;
+
+    /// Returns a short symbol or abbreviation for the NFT token/collection.
+    fn symbol(&self) -> String;
+
+    /// Returns the number of NFTs owned by a given account or contract
+    ///
+    /// @param owner - The account to query.
```
davidatwhiletrue: The account or contract to query.

text/0095-nft-standard.md
```diff
@@ -0,0 +1,387 @@
...
+    /// Returns the owner of a specific NFT.
+    ///
+    /// @param token_id - The ID of the NFT.
+    /// @return Option<Key> - The owner if it exists, else None.
+    fn owner_of(&self, token_id: U256) -> Option<Key>;
+
+    /// Performs a recipient check and transfers the ownership of an NFT.
+    /// @dev Reverts unless `runtime::caller()` is the current owner, an authorised
+    ///  operator, or the approved spender for this NFT. Reverts if `from` is not
+    ///  the current owner, if `to` is the default (zero‑hash) account, or if
```
davidatwhiletrue: (zero-hash) key ?

davidatwhiletrue (commented):
@Volodymyr-Kuchinskyi @ihor , please check my comments.

text/0095-nft-standard.md
Volodymyr-Kuchinskyi: I would put `(zero-hash) address`

davidatwhiletrue: yes, address better than key. 👍🏼

davidatwhiletrue (approved):
Looks good to me.

zie1ony (approved):

ihor (approved):

davidatwhiletrue (approved):
LGTM

mssteuer (approved):

PR #97, opened 2025-05-14
Update to CEP-18 Standard

zie1ony:
[rendered](https://github.com/odradev/ceps/blob/master/text/0018-token-standard.md)

Following update is the conclusion of all the conversations about token standards.
This is a more detailed and cleaned up version of the CEP-18 standard.

### DIFF
--- a/text/0018-token-standard.md
+++ b/text/0018-token-standard.md
@@ -1,251 +1,262 @@
 # Casper Payments: Casper Token Standard
 
 ## Summary
-A standard interface for the CSPR and the custom made tokens.
-
-## Motivation
-Having a token standard enables other projects to generalized on top it and makes
-integration easy, especially for wallets, DAOs and DeFi products like decentralized
-exchanges and liquidity pools. Defining a token standard before releasing
-the mainnet gives use the unique opportunity to implement this standard for our
-native platform token CSPR. It allows developers to reduce amount of code, 
-because it is no longer required to handle 2 cases: one for native
-token and the other for a tokens implementing current token standard. 
-
-In the design we decided to get away from the purse based model and use 
-the account based model. As a main inspiration we point at ERC-20 standard
-known from Ethereum. It is the most widely adopted standard in the blockchain
-space, successfully ported outside of Ethereum. Thanks to the unique ability
-of Casper to execute code in the account's context, the main drawback of ERC-20,
-which is the "approve and call" pattern is no longer a problem.
-
-## Guide-level explanation
-### Interface
-The token contract should implement following endpoints.
-
-#### name
-Returns the token name.
-```rust
-fn name() -> String
-```
-
-#### symbol
-Returns the token symbol.
-```rust
-fn symbol() -> String
-```
-
-#### decimals
-Returns the number of token decimals.
-```rust
-fn decimals() -> u8
-```
-
-#### balance_of
-Returns the amount of tokens given address holds.
-```rust
-fn balanceOf(address: Address) -> U512
-```
-
-#### transfer
-Transfer tokens from the direct function caller to the `recipient`.
-```rust
-fn transfer(recipient: Address, amount: U512)
-```
-
-#### approve
-Allow other address to transfer caller's tokens.
-```rust
-fn approve(spender: Address, amount: U512)
-```
-
-#### increase_allowance
-Atomically increases the allowance granted to spender by the caller.
-```rust
-fn increase_allowance(spender: Address, amount: U512)
-```
-
-#### decrease_allowance
-Atomically decreases the allowance granted to spender by the caller.
-```rust
-fn decrease_allowance(spender: Address, amount: U512)
-```
-
-#### allowance
-Returns the amount allowed to spend.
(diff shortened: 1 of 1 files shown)


text/0018-token-standard.md
```diff
@@ -1,251 +1,261 @@
...
+    fn decrease_allowance(&mut self, spender: Key, decr_by: U256);
+
+    /// Atomically increases the allowance granted to spender by the caller.
+    /// If the sum of the current allowance and `inc_by` is greater than the
+    /// maximum value of U256, the allowance is set to the maximum value of U256.
+    /// The `spender` is the address of the account that is allowed to spend
+    /// the tokens on behalf of the caller.
+    /// The `inc_by` is the amount by which the allowance should be increased.
+    fn increase_allowance(&mut self, spender: Key, inc_by: U256);
+}
```
gRoussac, line 91: Hi,

I would add `mint` and `burn` even if they are behind a configuration.

text/0018-token-standard.md
```diff
@@ -1,251 +1,261 @@
...
+when the appropriate conditions are met:
+
+```rust
+pub enum CEP18Error {
+    /// Spender does not have enough balance.
+    InsufficientBalance = 60001,
+    /// Spender does not have enough allowance approved.
+    InsufficientAllowance = 60002,
+    /// The user cannot target themselves.
+    CannotTargetSelfUser = 60003,
```
gRoussac, line 188: Right now this first errors in current implmentation are

```
pub enum Cep18Error {
    InvalidContext = 60000,
    InsufficientBalance = 60001,
    InsufficientAllowance = 60002,
    Overflow = 60003,
    PackageHashMissing = 60004,
```
    etc

text/0018-token-standard.md
```diff
@@ -1,251 +1,261 @@
...
+    hex::encode(hash_bytes)
+}
+```
 
-The standard doesn't cover the upgradeability story. Maybe it should?
+## Exisiting implementations
 
+The CEP-18 token implementations are available under the following links:
+- https://github.com/casper-ecosystem/cep18
+- https://github.com/odradev/odra/blob/HEAD/modules/src/cep18_token.rs
```
gRoussac, line 261: It seems to me strange to establish an interface using Casper Key type and having an implementation that differs by implementing the wrapper Address like 

```
pub fn transfer(&mut self, recipient: &Address, amount: &U256) {
```

It's a small difference but it kind of defeats the intention of this document somehow.

cc @darthsiroftardis

text/0018-token-standard.md
zie1ony: Most of the times during the development having `mint` and `burn` just made things unnecessarily complicated, and is not useful at all, as burning and minting logic is different from project to project. I'm strongly against having them in the standard.

Having it in `casper-ecosystem`'s implementation with modalities is still good idea as it serves as a standalone token with basic governance. But these doesn't have to be part of the standard.

text/0018-token-standard.md
zie1ony: The errors I propose are implementation-agnostic and comes from the logic of the token mechanics. Things like `InvalidContext` are framework-releated and shouldn't be here. In the `casper-ecosystem/cep18` implementation, you have errors that doesn't happen in Odra's implementation and vice versa.

text/0018-token-standard.md
zie1ony: Odra's Address serializes and deserializes to `Key`, but only supports Key's variants that makes sense. You can think of it as a implementation detail.

From Odra's codebase
```rust
impl CLTyped for Address {
    fn cl_type() -> CLType {
        CLType::Key
    }
}

impl ToBytes for Address {
    fn to_bytes(&self) -> Result<Vec<u8>, bytesrepr::Error> {
        Key::from(*self).to_bytes()
    }

    fn serialized_length(&self) -> usize {
        Key::from(*self).serialized_length()
    }
}

impl FromBytes for Address {
    fn from_bytes(bytes: &[u8]) -> Result<(Self, &[u8]), bytesrepr::Error> {
        let (key, remainder) = Key::from_bytes(bytes)?;

        let address = match key {
            Key::Account(account_hash) => Address::Account(account_hash),
            Key::Hash(raw_contract_package_hash) => {
                Address::Contract(ContractPackageHash::new(raw_contract_package_hash))
            }
            _ => return Err(bytesrepr::Error::Formatting)
        };

        Ok((address, remainder))
    }
}
```

While `casper-ecosystem` implementation actually allows for all kinds of Key types, but it still uses only AccountHash and PackageHash (and EntityAddr, but this most likely will never be turned on, more here: https://github.com/casper-network/casper-node/issues/4511). Other kinds of Keys simply doesn't make any sense to be recipient of tokens, as they would be unrecoverable.

Given all the above, I think `Key` as address is technically the right choice for the standard, from the lack of better choice.

zie1ony (commented):
@gRoussac thanks for reviewing

mssteuer (approved):

ihor (approved):

PR #99, opened 2026-05-12
CEP-2612: Permit Extension

zie1ony:
[Rendered][1]

[1]: https://github.com/odradev/ceps/blob/97cc79ec5f9bffdd96f573c1b5e82f30b4679ba9/text/2612-permit-extension.md

### DIFF
--- a/text/2612-permit-extension.md
+++ b/text/2612-permit-extension.md
@@ -147,12 +147,17 @@ The final digest is computed using the standard EIP-712 rule
 
 The `domainSeparator` is defined as:
 - `name` — the CEP-18 token's `name`,
-- `chainId` — the contract's `chain_name` (see Storage below); It should be 
+- `version` - the current major version of the signing domain,
+- `chain_name` — the contract's `chain_name` (see Storage below); It should be 
    in the [CAIP-2](https://github.com/ChainAgnostic/namespaces/blob/main/casper/caip2.md)
    format,
-- `verifyingContract` — the contract's own address.
+- `contract_package_hash` — the contract's own address.
 
-The salt and version fields are not used.
+Using `EIP-712` notation, the domain separator type string is:
+
+```
+EIP712Domain(string name,string version,string chain_name,bytes32 contract_package_hash)
+```
 
 ### Error codes
 
--- a/text/2612-permit-extension.md
+++ b/text/2612-permit-extension.md
@@ -24,8 +24,8 @@ Casper](https://github.com/casper-ecosystem/casper-eip-712).
 ## Specification
 
 The CEP-2612 extension is defined by:
-- one new entrypoint (`permit`),
-- the EIP-712 typed data definition for the `Permit` struct,
+- one new entry point (`permit`),
+- the EIP-712 typed data definition for the `Permit` struct and its domain separator,
 - error codes,
 - storage structure for replay-protection nonces and the EIP-712 chain
   identifier.
@@ -36,7 +36,7 @@ events are introduced by this CEP.
 
 Below definitions use Rust syntax, but they are not Rust specific.
 
-### Entrypoint interface
+### Entry point interface
 
 Contracts implementing this standard must expose the following entry point
 in addition to the CEP-18 interface:
@@ -55,7 +55,7 @@ pub trait CEP2612Interface {
     ///   disables the expiry check.
     /// - `public_key` is the signer's Casper public key.
     /// - `signature` is `public_key`'s signature over the EIP-712 digest
-    ///   of the `Permit` typed data (see below).
+    ///   of the `Permit` typed data.
     ///
     /// On success:
     /// - the allowance of `spender` over `owner` is set to `value` (i.e.
@@ -67,10 +67,12 @@ pub trait CEP2612Interface {
     /// The call must revert:
     /// - with `PermitExpired` if `deadline != u64::MAX` and the current
     ///   block time is strictly greater than `deadline`,
+    /// - with `InvalidPublicKey` if `owner` is not the `Address` derived
+    ///   from `public_key` (see "Signature scheme" below),
     /// - with `InvalidSignature` if `signature` does not verify against
     ///   `public_key` over the recomputed digest,
-    /// - with `InvalidSignature` if `owner` is not the `Address` derived
-    ///   from `public_key` (see "Signature scheme" below).
+    /// - with `InvalidNonce` if the `nonce` is not exactly equal to
+    ///   the one expected by the contract.
     ///
     /// The transaction caller (`env().caller()`) is irrelevant — `permit`
     /// is designed to be relayed by any third party.
@@ -98,10 +100,10 @@ the signer from the signature, so this CEP takes an explicit
 To preserve the property that a permit signature only authorizes its own
 `owner`'s allowance, implementations MUST enforce:
 
-1. `signature` is a valid signature of `public_key` over the EIP-712
-   digest of the `Permit` typed data (described below), AND
-2. `owner == Address::from(public_key)` — i.e. `owner` is the account
-   address corresponding to `public_key`.
+1. `owner == Address::from(public_key)` — i.e. `owner` is the account
+   address corresponding to `public_key` AND
+2. `signature` is a valid signature of `public_key` over the EIP-712
+   digest of the `Permit` typed data (described below).
 
 Without (2) a third party could sign a digest containing an unrelated
 `owner` field with their own key and have the contract set that
@@ -123,26 +125,30 @@ Its typehash is keccak256 of the type string above:
 PERMIT_TYPEHASH = 0x6e71edae12b1b97f4d1f60370fef10105fa2faae0126114a169c64845d6126c9
 ```
 
-The encoded message data hashed alongside the typehash is the
-concatenation, in order, of:
+`Address` encoding is defined as 33 bytes, where the first byte is a type tag:
+- `0x00` for `AccountHash`,
+- `0x01` for `PackageHash`.
 
-- `owner` encoded as an EIP-712 `address` (32 bytes, big-endian, left-padded),
-- `spender` encoded as an EIP-712 `address` (32 bytes, big-endian, left-padded),
-- `value` encoded as an EIP-712 `uint256` (32 bytes, big-endian),
-- `nonce` encoded as an EIP-712 `uint256` (32 bytes, big-endian) — equal
+The encoded message data hashed alongside the typehash is the
+concatenation of `CLValue` representations, in order, of:
+
+- `owner` encoded as an EIP-712 `address`,
+- `spender` encoded as an EIP-712 `address`,
+- `value` encoded as an EIP-712 `U256`,
+- `nonce` encoded as an EIP-712 `U256` — equal
   to the current on-chain `permit_nonces[owner]` value at the time the
   signature was produced,
-- `deadline` encoded as an EIP-712 `uint64` (32 bytes, big-endian,
+- `deadline` encoded as an EIP-712 `U256` (32 bytes, big-endian,
   left-padded).
 
 The final digest is computed using the standard EIP-712 rule
 (`keccak256("\x19\x01" || domainSeparator || keccak256(typeHash || encodedData))`).
 
-The EIP-712 `domainSeparator` uses:
+The `domainSeparator` is defined as:
 - `name` — the CEP-18 token's `name`,
-- `chainId` — the contract's `chain_name` (see Storage below); Casper
-  does not have a numeric chain id, so the chain name string is used in
-  its place,
+- `chainId` — the contract's `chain_name` (see Storage below); It should be 
+   in the [CAIP-2](https://github.com/ChainAgnostic/namespaces/blob/main/casper/caip2.md)
+   format,
 - `verifyingContract` — the contract's own address.
 
 The salt and version fields are not used.
@@ -155,12 +161,15 @@ codes when the appropriate conditions are met:
 ```rust
 pub enum CEP2612Error {
     /// Either `signature` does not verify against `public_key` over the
-    /// recomputed digest, or `owner` is not the account address derived
-    /// from `public_key`.
+    /// recomputed digest.
     InvalidSignature = 36_000,
     /// The current block time is strictly greater than `deadline`, and
     /// `deadline` is not the sentinel `u64::MAX`.
     PermitExpired = 36_001,
+    /// The `nonce` is not exactly equal to the one expected by the contract.
+    InvalidNonce = 36_002,
+    /// `owner` is not the `Address` derived from `public_key`.
+    InvalidPublicKey = 36_003,
 }
 ```
 
--- a/text/2612-permit-extension.md
+++ b/text/2612-permit-extension.md
@@ -105,9 +105,10 @@ To preserve the property that a permit signature only authorizes its own
 2. `signature` is a valid signature of `public_key` over the EIP-712
    digest of the `Permit` typed data (described below).
 
-Without (2) a third party could sign a digest containing an unrelated
-`owner` field with their own key and have the contract set that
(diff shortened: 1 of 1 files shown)


text/2612-permit-extension.md
```diff
@@ -0,0 +1,220 @@
...
+A token holder signs a typed message authorizing a `spender` to receive an
+allowance, and any party may submit that signature to the contract to set the
+allowance on the holder's behalf — enabling gas-less approvals for the holder.
+
+This CEP extends [CEP-18](0018-token-standard.md). It does not change CEP-18's
+`approve` / `allowance` / `transfer_from` semantics; it adds a new entry point
+that produces the same allowance state change as `approve` while being callable
+by a third party.
+
+## Prior art
```
davidatwhiletrue, line 16: Almost every CEP has a Motivation section. We may have one here too.

text/2612-permit-extension.md
```diff
@@ -0,0 +1,220 @@
...
+The signed payload follows [EIP-712](https://eips.ethereum.org/EIPS/eip-712)
+typed data hashing. EIP-712 is treated here as a dependency rather than
+re-specified — implementations are expected to use [EIP-712 toolkit for
+Casper](https://github.com/casper-ecosystem/casper-eip-712).
+
+## Specification
+
+The CEP-2612 extension is defined by:
+- one new entrypoint (`permit`),
+- the EIP-712 typed data definition for the `Permit` struct,
```
davidatwhiletrue: '...and the domain separator'

text/2612-permit-extension.md
```diff
@@ -0,0 +1,220 @@
...
+- storage structure for replay-protection nonces and the EIP-712 chain
+  identifier.
+
+The extension reuses CEP-18's `SetAllowance` event — `permit` produces the
+same allowance state change as `approve` and emits the same event. No new
+events are introduced by this CEP.
+
+Below definitions use Rust syntax, but they are not Rust specific.
+
+### Entrypoint interface
```
davidatwhiletrue: 'Entry point' with a space

text/2612-permit-extension.md
```diff
@@ -0,0 +1,220 @@
...
+    /// - `owner` is the token holder whose tokens are being approved.
+    /// - `spender` is the address that will be allowed to spend the tokens.
+    /// - `value` is the allowance amount.
+    /// - `deadline` is a block-time timestamp after which the permit is no
+    ///   longer accepted, expressed in the same units as the host
+    ///   environment's `get_block_time()`. The sentinel value `u64::MAX`
+    ///   disables the expiry check.
+    /// - `public_key` is the signer's Casper public key.
+    /// - `signature` is `public_key`'s signature over the EIP-712 digest
+    ///   of the `Permit` typed data (see below).
```
davidatwhiletrue: I'd remove `(see below)` becuase this is a code block.

text/2612-permit-extension.md
```diff
@@ -0,0 +1,220 @@
...
+    ///
+    /// On success:
+    /// - the allowance of `spender` over `owner` is set to `value` (i.e.
+    ///   the same effect as `owner` calling `approve(spender, value)`),
+    /// - the `owner`'s permit nonce is incremented by one,
+    /// - a CEP-18 `SetAllowance { owner, spender, allowance: value }`
```
davidatwhiletrue, line 64: just `SetAllowance { owner, spender, allowance }`

text/2612-permit-extension.md
```diff
@@ -0,0 +1,220 @@
...
+    ///   event is emitted.
+    ///
+    /// The call must revert:
+    /// - with `PermitExpired` if `deadline != u64::MAX` and the current
+    ///   block time is strictly greater than `deadline`,
+    /// - with `InvalidSignature` if `signature` does not verify against
+    ///   `public_key` over the recomputed digest,
+    /// - with `InvalidSignature` if `owner` is not the `Address` derived
```
davidatwhiletrue: why not returning `InvalidPublicKey`? I'd also reverse the order to match same comment in CEP-3009, and probably check logic in the implementation.

text/2612-permit-extension.md
```diff
@@ -0,0 +1,220 @@
...
+    ///   of the `Permit` typed data (see below).
+    ///
+    /// On success:
+    /// - the allowance of `spender` over `owner` is set to `value` (i.e.
+    ///   the same effect as `owner` calling `approve(spender, value)`),
+    /// - the `owner`'s permit nonce is incremented by one,
+    /// - a CEP-18 `SetAllowance { owner, spender, allowance: value }`
+    ///   event is emitted.
+    ///
+    /// The call must revert:
```
davidatwhiletrue, line 67: No means to know that a wrong `nonce` is the cause of a wrong signature.
This can be solved if, as suggested in another comment, we use random nonces as in CEP-3009.

text/2612-permit-extension.md
```diff
@@ -0,0 +1,220 @@
...
+#### Signature scheme — divergence from EIP-2612
+
+EIP-2612 passes the signature as `(v, r, s)` because Ethereum recovers the
+signer's address from the signature itself via ECDSA recovery and then
+asserts that the recovered address equals `owner`. Casper supports
+multiple signature schemes (ed25519 and secp256k1) and does not recover
+the signer from the signature, so this CEP takes an explicit
+`(public_key, signature)` pair instead.
+
+To preserve the property that a permit signature only authorizes its own
```
davidatwhiletrue, line 100: I would follow the same order as in CEP-3009. So, first check owner matches public key; then check signature. It's also how it should be implemented.

text/2612-permit-extension.md
```diff
@@ -0,0 +1,220 @@
...
+`owner`'s allowance, implementations MUST enforce:
+
+1. `signature` is a valid signature of `public_key` over the EIP-712
+   digest of the `Permit` typed data (described below), AND
+2. `owner == Address::from(public_key)` — i.e. `owner` is the account
+   address corresponding to `public_key`.
+
+Without (2) a third party could sign a digest containing an unrelated
+`owner` field with their own key and have the contract set that
+`owner`'s allowance. Both checks should revert with `InvalidSignature`.
```
davidatwhiletrue: Why not `InvalidPublicKey` for owner not matching signing public key?

text/2612-permit-extension.md
```diff
@@ -0,0 +1,220 @@
...
+multiple signature schemes (ed25519 and secp256k1) and does not recover
+the signer from the signature, so this CEP takes an explicit
+`(public_key, signature)` pair instead.
+
+To preserve the property that a permit signature only authorizes its own
+`owner`'s allowance, implementations MUST enforce:
+
+1. `signature` is a valid signature of `public_key` over the EIP-712
+   digest of the `Permit` typed data (described below), AND
+2. `owner == Address::from(public_key)` — i.e. `owner` is the account
```
davidatwhiletrue: Is `Address::from()` correct here?

text/2612-permit-extension.md
```diff
@@ -0,0 +1,220 @@
...
+Its typehash is keccak256 of the type string above:
+
+```
+PERMIT_TYPEHASH = 0x6e71edae12b1b97f4d1f60370fef10105fa2faae0126114a169c64845d6126c9
+```
+
+The encoded message data hashed alongside the typehash is the
+concatenation, in order, of:
+
+- `owner` encoded as an EIP-712 `address` (32 bytes, big-endian, left-padded),
```
davidatwhiletrue: In Casper, `owner` and `spender` are a hash of an Account Hash or a (package) Hash (the hash of 0x00 or 0x01 prefix || 32 bytes). I would mention it here. and remove big-endian, left-padded

text/2612-permit-extension.md
```diff
@@ -0,0 +1,220 @@
...
+- `spender` encoded as an EIP-712 `address` (32 bytes, big-endian, left-padded),
+- `value` encoded as an EIP-712 `uint256` (32 bytes, big-endian),
+- `nonce` encoded as an EIP-712 `uint256` (32 bytes, big-endian) — equal
+  to the current on-chain `permit_nonces[owner]` value at the time the
+  signature was produced,
+- `deadline` encoded as an EIP-712 `uint64` (32 bytes, big-endian,
```
davidatwhiletrue: “EIP-712 uint64" does not exist. u64 or uint256. let’s use the same here as in CEP-3009 for validAfter and validBefore

text/2612-permit-extension.md
```diff
@@ -0,0 +1,220 @@
...
+  left-padded).
+
+The final digest is computed using the standard EIP-712 rule
+(`keccak256("\x19\x01" || domainSeparator || keccak256(typeHash || encodedData))`).
+
+The EIP-712 `domainSeparator` uses:
```
davidatwhiletrue: Not aligned with casper-eip-712 domain type:

EIP712Domain(string name,string version,string chain_name,bytes32 contract_package_hash).

Worth to add the type string to this section too.

Also, recommend to use CAIP-2 for chain_name as [here](https://github.com/casper-ecosystem/casper-eip-712#casper-native-domain-example).

text/2612-permit-extension.md
```diff
@@ -0,0 +1,220 @@
...
+
+#### Simple values
+
+The chain name used in the EIP-712 domain separator is stored in the
+contract's named keys:
+
+- The chain name is stored under the key `chain_name` with the type
+  `String`.
+
+#### Permit nonces
```
davidatwhiletrue, line 200: Should we discuss non-sequential `nonce`s? I realize this will diverge from the ERC-2612 specification. But it’s a limitation and something Permit2 solved if I'm not mistaken.
Moving to random `nonce` probably means having a `cancel_permit`-like entry point, and a different structure to store used values.

davidatwhiletrue (commented):
@zie1ony , please check my comments.

text/2612-permit-extension.md
zie1ony: We are porting a standard here, not inventing anything new.

text/2612-permit-extension.md
zie1ony: done

text/2612-permit-extension.md
zie1ony: done

text/2612-permit-extension.md
zie1ony: done

text/2612-permit-extension.md
zie1ony: This is ok, because `value` is the parameter in `permit` function. This link makes sense.

text/2612-permit-extension.md
zie1ony: The main strategy in all the smart contracts we have is to revert on errors. And I'd like to keep it consistent.

I reordered it, but it doesn't really change anything.

text/2612-permit-extension.md
zie1ony: I added `InvalidNonce` error.

text/2612-permit-extension.md
zie1ony: Correct. I change the order.

text/2612-permit-extension.md
zie1ony: Done.

text/2612-permit-extension.md
zie1ony: Yes. It only represents the logic.

text/2612-permit-extension.md
zie1ony: done, I mentioned how address is serialized.

text/2612-permit-extension.md
zie1ony: Done

text/2612-permit-extension.md
zie1ony: I updated the document, but not sure if this is what you wanted. Please check

text/2612-permit-extension.md
zie1ony: I don't think we need to. We do 2612 to be compatible with possible 2612 users, but I'd use 3009 over 2612 in any case.

text/2612-permit-extension.md
davidatwhiletrue: I'm not convinced. `Address` is an Odra type. It doesn't exist in `casper_types`. The reader doesn't need to know that an Address is an account hash. It'd be more explicit.

text/2612-permit-extension.md
```diff
@@ -0,0 +1,229 @@
...
+- `nonce` encoded as an EIP-712 `U256` — equal
+  to the current on-chain `permit_nonces[owner]` value at the time the
+  signature was produced,
+- `deadline` encoded as an EIP-712 `U256` (32 bytes, big-endian,
+  left-padded).
+
+The final digest is computed using the standard EIP-712 rule
+(`keccak256("\x19\x01" || domainSeparator || keccak256(typeHash || encodedData))`).
+
+The `domainSeparator` is defined as:
```
davidatwhiletrue, line 148: This is not the domain separator for casper.
See here:
https://github.com/casper-ecosystem/casper-eip-712#casper-native-domain-example
```
EIP712Domain(string name,string version,string chain_name,bytes32 contract_package_hash)
```

text/2612-permit-extension.md
davidatwhiletrue: I don't see the domain type in the document. Because it differs from the EIP-2612 domain type, I think we need to include it.

```
EIP712Domain(string name,string version,string chain_name,bytes32 contract_package_hash)
```

Also, check this comment: https://github.com/casper-network/ceps/pull/99/changes#r3272351456

text/2612-permit-extension.md
```diff
@@ -0,0 +1,229 @@
...
+
+Its typehash is keccak256 of the type string above:
+
+```
+PERMIT_TYPEHASH = 0x6e71edae12b1b97f4d1f60370fef10105fa2faae0126114a169c64845d6126c9
+```
+
+`Address` encoding is defined as 33 bytes, where the first byte is a type tag:
+- `0x00` for `AccountHash`,
+- `0x01` for `PackageHash`.
```
davidatwhiletrue: `Hash` instead of `PackageHash`?
o "0x01 for a package `Hash`"

davidatwhiletrue (commented):
@zie1ony , see my comments.

mssteuer:
@zie1ony once @davidatwhiletrue's remaining comments are addressed/responded to, we can finalize this

text/2612-permit-extension.md
davidatwhiletrue: "Both checks should revert with `InvalidSignature`."
You need to update this sentence.

text/2612-permit-extension.md
```diff
@@ -0,0 +1,229 @@
...
+
+The encoded message data hashed alongside the typehash is the
+concatenation of `CLValue` representations, in order, of:
```
davidatwhiletrue: I would simplify this sentence to:
`"The encodeData value is the concatenation of the following EIP-712-encoded fields, in order:"`

'the concatenation of `CLValue` representations' is not correct.

I'd use `uint256` instead of `U256`.

text/2612-permit-extension.md
davidatwhiletrue: fixed in the last update

text/2612-permit-extension.md
davidatwhiletrue: discussed with Maciej. We keep it as it is.

text/2612-permit-extension.md
davidatwhiletrue: Fixed in the last update

text/2612-permit-extension.md
davidatwhiletrue: Fixed in the last update

mssteuer (approved):

davidatwhiletrue (approved):
LGTM

